PR #11, opened 2022-09-07
Improve CI/CD setup

danielgranhao:
This includes:

- A new GitHub Action that deploys the library into:
   - A swift package in [this repo](https://github.com/getlipa/lipa-lightning-lib-swift)
   - An Android library in [this repo](https://github.com/getlipa/lipa-lightning-lib-android). The Android library is built and delivered using JitPack
- Fixing a bug on the iOS native logger
- Setting up the existing Built & Test job to also build for the mobile targets

The deployment job runs anytime a tag is created with the naming `v*.*.*` and creates releases in the platform-specific repositories with that same name. An initial test release has been made to test the job.


.github/workflows/deploy-release.yml
```diff
@@ -0,0 +1,238 @@
+name: Deploy a new release
+
+on:
+  push:
+    tags:
+      - 'v*.*.*'
+
+jobs:
+  cross-compile-ios:
```
andrei-21, line 9:
```suggestion
  build-ios:
```
`cross` is confusing because below it says `use-cross: false`.
We can name the job the same way it is named in `build.yml`.

.github/workflows/build.yml
```diff
@@ -19,13 +19,47 @@ jobs:
...
+          toolchain: stable
+          profile: minimal
+          override: true
+          target: ${{ matrix.target }}
+      - name: Build target
+        uses: actions-rs/cargo@v1
+        with:
+          use-cross: false
+          command: build
+          args: --release --target ${{ matrix.target }}
```
andrei-21, line 43: Why `release`? Maybe `debug` if we are not going to use the artifacts?

.github/workflows/build.yml
```diff
@@ -19,13 +19,47 @@ jobs:
...
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+          target: ${{ matrix.target }}
+      - name: Build target
+        uses: actions-rs/cargo@v1
+        with:
+          use-cross: false
+          command: build
```
andrei-21, line 42: Shouldn't we add `test` step to make sure that we test the code for target platforms?

.github/workflows/build.yml
danielgranhao: Fixed!

.github/workflows/build.yml
danielgranhao: We generally can't because the binaries are being cross-compiled for other targets/architectures.

.github/workflows/deploy-release.yml
danielgranhao: The naming is still accurate because we are still cross-compiling even if not using the tool `cross` to do so.

I changed the iOS runner to ubuntu and enabled `cross` to maintain consistency. I previously experimented with not using `cross` to try to solve a bug that had nothing to do with this.

andrei-21 (approved):

PR #12, opened 2022-09-08
Fix missing `.ldk` folder error on example node

danielgranhao:
This solves the issue that if there's no `.ldk` folder on the root dir of the project, then the example node panics.

### DIFF
--- a/examples/node/main.rs
+++ b/examples/node/main.rs
@@ -169,6 +169,7 @@ fn main() {
 }
 
 fn init_logger() {
+    fs::create_dir_all(".ldk").unwrap();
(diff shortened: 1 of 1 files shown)


examples/node/main.rs
```diff
@@ -169,6 +169,7 @@ fn main() {
 }
 
 fn init_logger() {
+    fs::create_dir_all(".ldk").unwrap();
```
andrei-21, line 172: I would move this line above `init_logger();` since it is not only about logger initialization, but a directory for other things.

andrei-21 (approved):

PR #16, opened 2022-09-13
Init keys manager

andrei-21:


danielgranhao:
LGTM

I would only add a way for the host to get the mnemonic. It will be needed for backup. I see the following options:
- Add a new public method that converts a seed to a mnemonic.
- Return both the seed and mnemonic when creating a seed
- Return only the mnemonic and use the mnemonic for node initialisation

andrei-21:
> LGTM
> 
> I would only add a way for the host to get the mnemonic. It will be needed for backup. I see the following alternatives:
> 
>     * Add a new public method that converts a seed to a mnemonic.
> 
>     * Return both the seed and mnemonic when creating a seed
> 
>     * Return only the mnemonic and use the mnemonic for node initialisation

Good point! I will work on this.

src/secret.rs
```diff
@@ -0,0 +1,19 @@
+/// An object that holds nessecary secret.
```
danielgranhao, line 1:
```suggestion
/// An object that holds necessary secrets.
```

danielgranhao (approved):

PR #21, opened 2022-09-14
Init persister

danielgranhao:
This assumes that we use the same persistence callback for `ChannelMonitors` and the other objects that need to be persisted (e.g. `ChannelManager`, `NetworkGraph`, ...).

### DIFF
--- a/src/storage_persister.rs
+++ b/src/storage_persister.rs
@@ -62,7 +62,6 @@ impl StoragePersister {
 
     fn persist_object(&self, bucket: String, key: String, data: Vec<u8>) -> Result<(), Error> {
         if !self.storage.put_object(bucket, key, data) {
-            // Operating System I/O Error
             return Err(Error::new(
                 io::ErrorKind::Other,
                 "Failed to persist object using storage callback",
--- a/src/storage_persister.rs
+++ b/src/storage_persister.rs
@@ -40,7 +40,7 @@ impl StoragePersister {
     }
 
     pub fn check_object_storage_health(&self) -> bool {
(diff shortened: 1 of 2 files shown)


src/storage_persister.rs
```diff
@@ -25,9 +39,33 @@ impl StoragePersister {
...
+    }
+
+    pub fn read_scorer(&self) {
+        // TODO: Implement
+    }
+
+    fn persist_object(&self, bucket: String, key: String, data: Vec<u8>) -> Result<(), Error> {
+        if !self.storage.put_object(bucket, key, data) {
+            // Operating System I/O Error
+            return Err(Error::from_raw_os_error(5));
```
andrei-21, line 65: Can we explicitly create `Error` with a value from [`ErrorKind` enum](https://doc.rust-lang.org/std/io/enum.ErrorKind.html) and a message (using `Error::new()` or `Error::other()` method)?

src/storage_persister.rs
```diff
@@ -1,16 +1,30 @@
...
+use lightning::util::logger::Logger;
+use lightning::util::persist::Persister;
 use lightning::util::ser::Writeable;
+use std::io::Error;
+use std::ops::Deref;
 
 use crate::callbacks::RedundantStorageCallback;
 
 static MONITORS_BUCKET: &str = "monitors";
+static OBJECT_BUCKET: &str = "objects";
```
andrei-21, line 23: If we follow monitors bucket naming:
```suggestion
static OBJECTS_BUCKET: &str = "objects";
```

andrei-21 (approved):

PR #22, opened 2022-09-14
Read channel monitors

andrei-21:


danielgranhao (commented):
LGTM if we always get the `ChannelMonitor` objects from local storage. How will we handle the situation in which we want to use the monitors persisted remotely?

andrei-21:
> LGTM if we always get the `ChannelMonitor` objects from local storage. How will we handle the situation in which we want to use the monitors persisted remotely?

We get objects using `redundant_storage_callback`, so it is their business to make sure that they serve the latest objects.
We should never explicitly request local or remote version.

danielgranhao (approved):

PR #27, opened 2022-09-16
Implement async runtime construction

andrei-21:


src/async_runtime.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    pub fn spawn<F>(&self, future: F) -> JoinHandle<F::Output>
+    where
+        F: Future + Send + 'static,
+        F::Output: Send + 'static,
+    {
+        self.rt.spawn(future)
+    }
+
+    pub fn spawn_regularly<Func, F>(&self, duration: Duration, func: Func) -> JoinHandle<()>
```
danielgranhao, line 34: I think this name is a bit confusing. It makes it seem like `spawn` will be called regularly but it's called only once. Not sure about the following suggestion either, but I think it's an improvement.

```suggestion
    pub fn spawn_repeating_task<Func, F>(&self, duration: Duration, func: Func) -> JoinHandle<()>
```

src/async_runtime.rs
```diff
@@ -0,0 +1,109 @@
...
+
+pub struct AsyncRuntime {
+    rt: Runtime,
+}
+
+impl AsyncRuntime {
+    pub fn new() -> Result<Self, InitializationError> {
+        let rt = Builder::new_multi_thread()
+            .worker_threads(4)
+            .thread_name("3l-asyn-runtime")
```
danielgranhao, line 17:
```suggestion
            .thread_name("3l-async-runtime")
```

src/lib.rs
```diff
@@ -31,6 +33,8 @@ impl LightningNode {
         config: Config,
         redundant_storage_callback: Box<dyn RedundantStorageCallback>,
     ) -> Result<Self, InitializationError> {
+        let _rt = AsyncRuntime::new()?;
```
danielgranhao, line 36: We could already store the runtime in `LightningNode`. Otherwise, the runtime is dropped when `new()` returns.

danielgranhao (approved):

PR #28, opened 2022-09-16
Update README.md to use Nigiri environment

gcomte:
So far there is no network environment being used by the lipa-lightning-library, so this PR only changes the README.md without any effects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,10 +7,12 @@
 
 # Test
 
-Start bitcoin core:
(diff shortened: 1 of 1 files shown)


gcomte:
The idea is to use Nigiri instead of Polar for 2 reasons:

- With Nigiri you can fire up an Esplora instance, which you can't do on Polar
- We need a headless tool for CI integration/tests

danielgranhao (approved):

PR #29, opened 2022-09-16
Add detached Esplora client

gcomte:
This adds an Esplora client to the project, which is however not being used yet by the library. That's why it declares `#[allow(dead_code)]`, which should be undone later, once the client is actually being used.

The client contains all the functionality required for keeping the chain synced for LDK. Not more not less. There is more functionality to be found in the linked source file, linked to in the top of the files `src/esplora_client.rs` and `src/esplora_client_api.rs`

Even though this PR introduces detached code, that is never being called, I still wanted to bring this code into discussion early, because there are a lot of design decisions that should be considered:

- Copied code: Should we rather rely on forks of forks of forks or copy the code and have our own implementation? Should we switch to use maintained libraries later, when the code has eventually been merged?
- Given the nature of copied code that might be replaced by using libraries later .. should we provide any testing?
- Is it okay to go for an `async` client or should we rather develop a `sync` client?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,8 @@ bitcoin = "0.28.1"
 lightning = { version = "0.0.110", features = ["max_level_trace"] }
 log = "0.4.17"
 rand = "0.8.5"
+reqwest = { version = "0.11", features = ["json", "socks"] }
+serde = "1.0.144"
 thiserror = "1.0.34"
 tokio = { version = "1.21.1", features = ["rt-multi-thread", "time"] }
 uniffi = "0.20.0"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -13,6 +13,8 @@ bitcoin = "0.28.1"
 lightning = { version = "0.0.110", features = ["max_level_trace"] }
 log = "0.4.17"
 rand = "0.8.5"
+reqwest = { version = "0.11", features = ["json", "socks"] }
```
danielgranhao, line 16: Change the `reqwest` features so that we can build for other platforms. 

```suggestion
reqwest = { version = "0.11", default-features = false, features = ["json", "socks", "rustls-tls"] }
```

gcomte:
It seems that we can use the [rust-esplora-client](https://github.com/bitcoindevkit/rust-esplora-client) directly by now

PR #31, opened 2022-09-16
Implement tx broadcast interface

andrei-21:


src/lib.rs
```diff
@@ -21,28 +22,39 @@ use crate::keys_manager::{generate_secret, init_keys_manager};
...
+    esplora_client: Arc<AsyncClient>,
 }
 
 impl LightningNode {
     pub fn new(
         config: Config,
         redundant_storage_callback: Box<dyn RedundantStorageCallback>,
     ) -> Result<Self, InitializationError> {
-        let rt = AsyncRuntime::new()?;
+        let rt = Arc::new(AsyncRuntime::new()?);
```
danielgranhao, line 45: Instead of wrapping the runtime in an `Arc`, I think we can use a `tokio::runtime::Handle`.

danielgranhao (approved):
Only a minor detail. Otherwise LGTM.

danielgranhao:
I've created an [issue](https://github.com/bitcoindevkit/rust-esplora-client/issues/11) to see if they would be open to allow the selection of the TLS implementation used by `reqwest`.

Cargo.toml
```diff
@@ -17,8 +17,11 @@ thiserror = "1.0.34"
 tokio = { version = "1.21.1", features = ["rt-multi-thread", "time"] }
 uniffi = "0.20.0"
 
+esplora-client = { git = "https://github.com/bitcoindevkit/rust-esplora-client" }
```
gcomte, line 20: Specifying the async feature would safe us compilation time and create a smaller binary
`
esplora-client = { git = "https://github.com/bitcoindevkit/rust-esplora-client", default-features = false, features = ["async"] }`

gcomte:
Clippy throws warnings ... (Our clippy CI check isn't working as intended yet)

Cargo.toml
andrei-21: Good point, thanks.

src/lib.rs
andrei-21: You are right, I will address it in a separate PR.

andrei-21: https://github.com/getlipa/lipa-lightning-lib/pull/33

danielgranhao (approved):

PR #34, opened 2022-09-19
Fix Clippy warnings

gcomte:
Due to our clippy check CI not working as intended, some clippy warnings are now being thrown on develop.

This PR removes them, by mostly making the clippy rules less strict, which must be undone as soon as the different yet-to-be-developed parts of the project are finally integrated!

PR #35, opened 2022-09-19
Fail CI on clippy warnings

danielgranhao:
Configures Clippy CI job to fail if there are any warnings and fixes warnings. Warnings were fixed by ignoring certain Clippy rules because of the current software being a WIP.


gcomte:
For the record: https://github.com/getlipa/lipa-lightning-lib/pull/34#issuecomment-1250834093

gcomte (approved):

PR #37, opened 2022-09-19
Init ChainMonitor

andrei-21:
I wrapped some types to `Box` to satisfy `Deref` trait requirements. Later we can switch to `Arc` if we see a need for this.


andrei-21:
Closing in favor of https://github.com/getlipa/lipa-lightning-lib/pull/38.

PR #38, opened 2022-09-20
Init channel manager

andrei-21:


src/lib.rs
```diff
@@ -98,7 +121,31 @@ impl LightningNode {
...
         // Step 8. Initialize the ChannelManager
         let mobile_node_user_config = build_mobile_node_user_config();
-        let _channel_manager = persister.read_channel_manager(mobile_node_user_config);
+        // TODO: Init properly.
+        let best_block = BestBlock::from_genesis(config.network);
+        let chain_params = ChainParameters {
+            network: config.network,
+            best_block,
+        };
+        let mut_channle_monitors: Vec<&mut ChannelMonitor<InMemorySigner>> =
```
danielgranhao, line 130:
```suggestion
        let mut_channel_monitors: Vec<&mut ChannelMonitor<InMemorySigner>> =
```

src/storage_persister.rs
```diff
@@ -75,8 +76,59 @@ impl StoragePersister {
         result
     }
 
-    pub fn read_channel_manager(&self, _user_config: UserConfig) {
-        // TODO: Implement
+    #[allow(clippy::too_many_arguments, clippy::type_complexity)]
+    pub fn read_channel_manager<M, T, F, L>(
```
danielgranhao, line 80: I would prefer if this method only handled reading a preexisting `ChannelManager` and if not available, we create a fresh one in another place. Like it is now, we are putting initialisation logic in the `StoragePersister`, which I don't think is intuitive.

src/storage_persister.rs
```diff
@@ -75,8 +76,59 @@ impl StoragePersister {
...
+            let channel_manager = SimpleArcChannelManager::new(
+                fee_estimator,
+                chain_monitor,
+                broadcaster,
+                logger,
+                keys_manager,
+                user_config,
+                chain_params,
+            );
+            Ok((None, channel_manager))
```
danielgranhao, line 130: Why don't we return the blockhash if it's a new `ChannelManager`? We have it available in `chain_params.best_block.block_hash()`.

danielgranhao: Ok, I see that you are using this to know if this method is reading or creating a `ChannelManager`. Maybe we can move the "read or create" logic directly to `LightningNode::new()` or create a new function in `lib.rs` that handles that.

src/storage_persister.rs
andrei-21: I feel the same, but other ideas I tried look uglier.
What about we rename the method to `read_or_init_channel_manager()` to make it more explicit?

danielgranhao (approved):

PR #42, opened 2022-09-21
Init PeerManager

danielgranhao:


src/lib.rs
```diff
@@ -166,6 +179,22 @@ impl LightningNode {
         let _graph = persister.read_graph();
 
         // Step 12. Initialize the PeerManager
+        let mut ephemeral_bytes = [0; 32];
+        OsRng.try_fill_bytes(&mut ephemeral_bytes).map_err(|e| {
+            InitializationError::SecretGeneration {
+                message: e.to_string(),
+            }
+        })?;
```
gcomte, line 187: Should we move out the crypto stuff into its own module where it can be reused?

src/lib.rs
```diff
@@ -166,6 +179,22 @@ impl LightningNode {
         let _graph = persister.read_graph();
 
         // Step 12. Initialize the PeerManager
+        let mut ephemeral_bytes = [0; 32];
+        OsRng.try_fill_bytes(&mut ephemeral_bytes).map_err(|e| {
```
gcomte, line 183: Should we move out the crypto stuff into its own module where it can be reused?

gcomte (commented):
Minor discussion point, LGTM otherwise

src/lib.rs
danielgranhao: I was starting to do that and then realised that it wouldn't be that reusable due to the returned error. To make it reusable, we would still have to `map_err()` in both places this is used, and we would end up with almost the same repeated code.

andrei-21: I will have a look into it.

danielgranhao:
LGTM

gcomte (approved):

PR #43, opened 2022-09-21
Start background processing

danielgranhao:
RapidGossipSync and ProbabilisticScorer instances are created just to be fed to the background processor. Properly initializing them will be handled in later PRs.

The [docs](https://docs.rs/lightning-background-processor/0.0.110/lightning_background_processor/struct.BackgroundProcessor.html#data-persistence) state that a persistence error will make the background processor stop and that this should be handled by calling `join()`, handling the error, and finally starting the background processor again. 

To deal with that, instead of just having the initial thread launching the thread in which the background processor runs, we have a new intermediate thread that continually launches the background processor, joins, launches again, and so forth.

The main issue I'm experiencing is that now there's no easy way for the original thread to stop the background processor gracefully. For that to be possible, the new thread must not block on the `join()`. That would be possible if the `JoinHandle` field of `lightning_background_processor::BackgroundProcessor` was public. That way, the new thread could call `JoinHandle.is_finished()` and avoid getting stuck in `join()`.


danielgranhao:
In private discussions, it was decided to just assume that persistence calls made by the `BackgroundProcessor` don't fail. This simplifies this PR a lot and shouldn't be too hard to implement on the persistence layer.

src/lib.rs
```diff
@@ -178,32 +184,60 @@ impl LightningNode {
...
+            Arc::clone(&logger),
+        )));
 
         // Step 17. Initialize the InvoicePayer
 
         // Step 18. Initialize the Persister
         // Persister trait already implemented and instantiated ("persister")
 
         // Step 19. Start Background Processing
+        // This assumes that the background processor will never fail.
```
gcomte, line 219: Could you elaborate on what you mean by *this*?

Do you mean the fact that we just `start()` and then never join and restart on potential errors?

andrei-21: I elaborated in the comment.

gcomte (approved):

PR #44, opened 2022-09-22
Refactoring: Sort InitializationErrors alphabetically

gcomte:


andrei-21:
In udl file also then?

andrei-21 (approved):

PR #47, opened 2022-09-22
Switch to blocking esplora client

gcomte:


src/lib.rs
```diff
@@ -49,7 +47,6 @@ static ESPLORA_TIMEOUT_SECS: u64 = 30;
 #[allow(dead_code)]
 pub struct LightningNode {
     #[allow(dead_code)]
-    rt: AsyncRuntime,
```
andrei-21, line 52: **_Hold on, not so reckless!_** We will need the async runtime for the background processor.

gcomte: Oh. I force-pushed a new version, which purges a lot less aggressively.

andrei-21 (approved):

PR #52, opened 2022-09-26
Implement outgoing P2P connection to LSP

gcomte:
Open a connection to another Lightning node (LSP) over the LN P2P.

This is necessary, because we don't listen to incoming connections, so the only way to connect with the LSP is an outgoing connection.

### DIFF
--- a/src/chain_access.rs
+++ b/src/chain_access.rs
@@ -5,7 +5,7 @@ use lightning::chain::{Confirm, Filter, WatchedOutput};
 use std::sync::{Arc, Mutex};
 
 #[allow(dead_code)]
-pub(crate) struct LipaChainAccess {
+pub struct LipaChainAccess {
     esplora_client: Arc<BlockingClient>,
     queued_txs: Mutex<Vec<(Txid, Script)>>,
     watched_txs: Mutex<Vec<(Txid, Script)>>,
--- a/src/chain_access.rs
+++ b/src/chain_access.rs
@@ -5,7 +5,7 @@ use lightning::chain::{Confirm, Filter, WatchedOutput};
 use std::sync::{Arc, Mutex};
 
 #[allow(dead_code)]
-pub struct LipaChainAccess {
+pub(crate) struct LipaChainAccess {
     esplora_client: Arc<BlockingClient>,
     queued_txs: Mutex<Vec<(Txid, Script)>>,
     watched_txs: Mutex<Vec<(Txid, Script)>>,
--- a/src/chain_access.rs
+++ b/src/chain_access.rs
@@ -1,4 +1,4 @@
-use crate::errors::ChainSyncError;
+use crate::errors::RuntimeError;
 use bitcoin::{Script, Transaction, Txid};
 use esplora_client::blocking::BlockingClient;
 use lightning::chain::{Confirm, Filter, WatchedOutput};
@@ -32,7 +32,7 @@ impl LipaChainAccess {
     pub(crate) fn sync(
         &self,
         _confirmables: Vec<&(dyn Confirm + Sync)>,
-    ) -> Result<(), ChainSyncError> {
+    ) -> Result<(), RuntimeError> {
         // TODO: sync with the chain
         // this will include moving queued txs and outputs to watched ones
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -21,7 +21,7 @@ mod tx_broadcaster;
 use crate::async_runtime::AsyncRuntime;
 use crate::callbacks::RedundantStorageCallback;
 use crate::chain_access::LipaChainAccess;
-use crate::config::{Config, LspConfig};
+use crate::config::{Config, NodeAddress};
 use crate::errors::{InitializationError, RuntimeError};
 use crate::event_handler::LipaEventHandler;
 use crate::fee_estimator::FeeEstimator;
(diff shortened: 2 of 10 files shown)


src/lipalightninglib.udl
```diff
@@ -9,6 +9,12 @@ dictionary Config {
     Network network;
     sequence<u8> seed;
     string esplora_api_url;
+    LspConfig lsp;
+};
+
+dictionary LspConfig {
```
andrei-21, line 15:
```suggestion
dictionary NodeAddress {
```
But in `Config` we can have it as `NodeAddress lsp_node;`.

src/p2p_networking.rs
```diff
@@ -0,0 +1,52 @@
...
+use bitcoin::secp256k1::PublicKey;
+use log::debug;
+use std::net::SocketAddr;
+use std::sync::Arc;
+use std::task::Poll;
+use std::time::Duration;
+use tokio::time::sleep;
+
+pub fn connect_to_peer(
+    rt: &AsyncRuntime,
```
andrei-21, line 12:
```suggestion
    handle: &Handle,
```
We should have only one instance of the async runtime and pass handles to downstream functions.

andrei-21: Moreover let's make this function `async`, let the caller to block on it if needed.

src/p2p_networking.rs
```diff
@@ -0,0 +1,52 @@
...
+    peer_manager: Arc<PeerManager>,
+    pubkey: PublicKey,
+    addr: SocketAddr,
+) -> Result<(), RuntimeError> {
+    rt.handle().block_on(async move {
+        let result =
+            lightning_net_tokio::connect_outbound(Arc::clone(&peer_manager), pubkey, addr).await;
+
+        if let Some(connection_closed_future) = result {
+            let mut connection_closed_future = Box::pin(connection_closed_future);
```
andrei-21, line 22: Why do we need to pin here?

andrei-21 (changes requested):

src/tx_broadcaster.rs
```diff
@@ -4,7 +4,7 @@ use lightning::chain::chaininterface::BroadcasterInterface;
 use log::error;
 use std::sync::Arc;
 
-pub(crate) struct TxBroadcaster {
```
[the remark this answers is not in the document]
gcomte: It was required because of `connect_to_peer()` being `pub` as well. But now that I'm not exposing `connect_to_peer()` anymore, I'll revert this change.

src/p2p_networking.rs
gcomte: I think it's because `poll(future)` [requires its param to implement the Unpin trait](https://github.com/rust-lang/futures-rs/blob/7a98cf0bbeb397dcfaf5f020b371ab9e836d33d4/futures-util/src/async_await/poll.rs#L24).

gcomte: https://rust-lang.github.io/async-book/04_pinning/01_chapter.html

andrei-21: Thanks!

src/lib.rs
```diff
@@ -197,6 +200,14 @@ impl LightningNode {
         )?);
 
         // Step 13. Initialize Networking
+        let peer_mgr = Arc::clone(&peer_manager);
+        rt.handle().block_on(async move {
+            P2pConnections::connect_peer(&config.lsp_node, Arc::clone(&peer_mgr))
```
andrei-21, line 205:
```suggestion
        let peer_manager_clone = Arc::clone(&peer_manager);
        rt.handle().block_on(async move {
            P2pConnections::connect_peer(&config.lsp_node, peer_manager_clone)
```

gcomte: Fixed here: https://github.com/getlipa/lipa-lightning-lib/pull/52/commits/5150ee854808d0440586cf8928c7c4711ba871df

src/p2p_networking.rs
gcomte: Fixed. It is now being `blocked_on` in the lib.rs initialization flow.

src/tx_broadcaster.rs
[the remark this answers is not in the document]
gcomte: Fixed.

src/lib.rs
```diff
@@ -197,6 +200,15 @@ impl LightningNode {
         )?);
 
         // Step 13. Initialize Networking
+        let peer_manager_clone = Arc::clone(&peer_manager);
+
+        rt.handle().block_on(async move {
+            P2pConnections::connect_peer(&config.lsp_node, Arc::clone(&peer_manager_clone))
```
andrei-21, line 206:
```suggestion
            P2pConnections::connect_peer(&config.lsp_node, peer_manager_clone)
```

danielgranhao (approved):

src/chain_access.rs
```diff
@@ -1,11 +1,11 @@
-use crate::errors::ChainSyncError;
+use crate::errors::RuntimeError;
 use bitcoin::{Script, Transaction, Txid};
 use esplora_client::blocking::BlockingClient;
 use lightning::chain::{Confirm, Filter, WatchedOutput};
 use std::sync::{Arc, Mutex};
 
 #[allow(dead_code)]
-pub(crate) struct LipaChainAccess {
```
andrei-21, line 8: We should revert `pub(crate)` here and below, shouldn't we?

gcomte: Absolutely, thanks! https://github.com/getlipa/lipa-lightning-lib/pull/52/commits/e874b97993063e56f6dcc17df7f658bb0fc7bb28

andrei-21 (approved):

PR #57, opened 2022-10-03
Setup integration test suite with nigiri

gcomte:


andrei-21 (approved):
Yes... having a crate just for one struct looks like an overkill, but I cannot think about a better solution. Maybe in the future we will have more code we want to share and it will be normal then.

PR #72, opened 2022-10-19
Add lpsd submodule and client

andrei-21:


src/callbacks.rs
```diff
@@ -28,6 +28,16 @@ pub trait RedundantStorageCallback: Send + Sync + Debug {
...
 }
+
+pub trait LspCallback: Send + Sync {
+    fn channel_information(&self) -> Vec<u8>;
+
+    /// Register a new incoming payment.
+    ///
+    /// # Return
+    /// Returns non empty string with description in case of an error.
+    fn register_payment(&self, bytes: Vec<u8>) -> String;
```
danielgranhao, line 42: We can now handle exceptions with the new version of UniFFI 😃 see [this](https://mozilla.github.io/uniffi-rs/udl/callback_interfaces.html). We should fix this in other callbacks as well.

lspd/Makefile
```diff
@@ -4,17 +4,18 @@ LND_DATA = ${NIGIRI_DATA}/volumes/lnd
 all: lnd.env
 
 lnd.env: ${LND_DATA}/tls.cert ${LND_DATA}/data/chain/bitcoin/regtest/admin.macaroon
-	rm --force $@
 	@echo "Generating $@"
-	@echo -n "LND_CERT=" > $@
-	@sed --null-data 's/\n/\\n/g' ${LND_DATA}/tls.cert >> $@
+        # Explicitly using /bin/echo to deal with `-n`, otherwise it does not work on MaxOS
```
danielgranhao, line 8:
```suggestion
        # Explicitly using /bin/echo to deal with `-n`, otherwise it does not work on macOS
```

examples/node/main.rs
```diff
@@ -18,24 +26,33 @@ static BASE_DIR: &str = ".ldk";
...
+    let lsp_info = ChannelInformationReply::decode(&*lsp_info).unwrap();
+    let ln_node_address = NodeAddress {
+        pub_key: lsp_info.pubkey,
+        address: lsp_info.host,
+    };
+    info!("Lsp pubkey: {}", lsp_info.lsp_pubkey.to_hex());
+    info!("LN node {:?}", ln_node_address);
 
     // Create dir for node data persistence.
     fs::create_dir_all(BASE_DIR).unwrap();
```
danielgranhao, line 47: Need to move this to the beginning of `main()` otherwise the app crashes when `BASE_DIR` doesn't exist

danielgranhao (approved):

PR #75, opened 2022-10-20
Implement and test graceful shutdown of background tasks

andrei-21:


danielgranhao (approved):

danielgranhao (approved):

src/async_runtime.rs
```diff
@@ -29,43 +30,86 @@ impl AsyncRuntime {
...
-        self.handle.spawn(async move {
+        let (stop_sender, mut stop_receiver) = mpsc::channel(1);
+        let (status_sender, status_receiver) = mpsc::channel(1);
+        let handle = self.handle.spawn(async move {
             let mut interval = time::interval(interval);
             interval.set_missed_tick_behavior(time::MissedTickBehavior::Skip);
+            interval.tick().await;
+
+            // Moved here so that it's dropped when the task finishes
+            let _status_sender = status_sender;
```
gcomte, line 67: What about just adding `drop(status_sender);` at the end of the task?

danielgranhao: Yes, thanks for the suggestion! Fixed in 058f2ed.

src/async_runtime.rs
```diff
@@ -29,44 +30,83 @@ impl AsyncRuntime {
...
+}
+
+impl RepeatingTaskHandle {
+    pub(crate) fn safe_abort(&self) {
+        self.stop_sender.blocking_send(()).unwrap();
+    }
+
+    pub(crate) fn block_until_finished(&mut self) {
+        self.status_receiver.blocking_recv();
+    }
```
gcomte, line 101: What's the reason between separating `safe_abort()` and `block_until_finished`? Is there any use case for that, or would it make more sense to make `safe_abort()` blocking until finished itself? Or maybe add a third function that calls both of these functions and is called `safe_abort_blocking()` or something?

As a programmer using these functions I might think `safe_abort()` is itself already blocking and forget to `block_until_finished()`.

danielgranhao: I see your point. Right now, a blocking version would be enough, but I think there might be use cases in which we don't care in which order two or more tasks are stopped. In that case, it's more efficient to call`safe_abort()` on them and only wait after.

> As a programmer using these functions I might think safe_abort() is itself already blocking and forget to block_until_finished().

I think of `safe_abort()` as a variation of the original `JoinHandle<>.abort()`, and that is my main argument for not making it blocking.

I added the `safe_abort_blocking()` method in bb46507.

src/async_runtime.rs
```diff
@@ -29,43 +30,86 @@ impl AsyncRuntime {
...
 }
 
+pub(crate) struct RepeatingTaskHandle {
+    handle: JoinHandle<()>,
+    stop_sender: mpsc::Sender<()>,
+    status_receiver: mpsc::Receiver<()>,
+}
+
+impl RepeatingTaskHandle {
+    pub(crate) fn safe_abort(&self) {
```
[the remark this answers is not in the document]
danielgranhao: I renamed it to `start_shutdown` to imply that it's a version that doesn't wait for the shutdown to complete. The new blocking version became `shutdown_blocking`. Could also be simply `shutdown` but this way I think it's more obvious.

gcomte (approved):

PR #79, opened 2022-10-24
Fix dependencies

andrei-21:
Fix dependencies mess. We should be careful to use `secp256k1`, `hashes` from `bitcoin`, otherwise we risk to get incompatible structures from different versions of the same crate.

### DIFF
--- a/src/encryption.rs
+++ b/src/encryption.rs
@@ -1,9 +1,9 @@
 use crate::errors::RuntimeError;
 
 use aes::cipher::{block_padding::Pkcs7, BlockEncryptMut, BlockSizeUser, KeyIvInit};
-use bdk::bitcoin::secp256k1::rand::thread_rng;
-use bdk::bitcoin::secp256k1::{PublicKey, Secp256k1, SecretKey};
 use bitcoin::hashes::{sha256, sha512, Hash, HashEngine, Hmac, HmacEngine};
+use bitcoin::secp256k1::scalar::Scalar;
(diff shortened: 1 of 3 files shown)


src/encryption.rs
```diff
@@ -16,9 +16,9 @@ type Aes256CbcEnc = cbc::Encryptor<aes::Aes256>;
 // https://pkg.go.dev/github.com/btcsuite/btcd/btcec#Encrypt
 // https://github.com/btcsuite/btcd/blob/v0.22.1/btcec/ciphering.go#L70
 #[allow(dead_code)]
-pub(crate) fn encrypt(pubkey: PublicKey, data: &[u8]) -> Result<Vec<u8>, RuntimeError> {
-    let secp = Secp256k1::new();
-    let (ephemeral, ephemeral_pubkey) = secp.generate_keypair(&mut thread_rng());
+pub(crate) fn encrypt(pubkey: &PublicKey, data: &[u8]) -> Result<Vec<u8>, RuntimeError> {
+    let secp = secp256k1::Secp256k1::new();
```
gcomte, line 20:
```suggestion
    let secp = Secp256k1::new();
```

gcomte: Why did you change this? Imo not necessary.

gcomte: Oh okay, sorry for the false alarm.

gcomte (approved):

PR #81, opened 2022-10-25
Add error

andrei-21:
I find our error model very detailed, but useless. It precisely says what has happened, but does not give any clue how to react on it. I suggest we introduce only one generic `Error` enum and use it where we want to return an error in internal or external code.
Some functions will help us convert io/LDK error into our types like:
```rust
fn foo() -> Result<(), Error> {
    // Result into -> InvalidInput: <message from parse()>
    let request = parse().map_err(invalid_input)?;

    // Result into -> RuntimeError: Email client error: <message from call()>
    let response = email_client.call(request).map_err(runtime_error_with("Email client error")?;

    // If InvalidInput, result into -> PermanentFailure: InvalidInput: <message from our_internal_function()>
    // else -> <message from our_internal_function()>
    our_internal_function(response).map_err(lift_invalid_input)?
}

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,5 +1,104 @@
 use uniffi::ffi::foreigncallbacks::UnexpectedUniFFICallbackError;
 
+#[derive(Debug, PartialEq, Eq, thiserror::Error)]
+pub enum Error {
+    /// Invalid input.
+    /// Consider fixing the input and retrying the request.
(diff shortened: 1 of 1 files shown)


danielgranhao (approved):
Cool idea! We might in some scenarios still want more specific errors so that it's easier to react differently to different exception in the mobile side without looking into the message. We'll figure out later if that's needed or not.

PR #82, opened 2022-10-25
Test sync after restart + bug fixes

danielgranhao:


tests/chain_sync_test.rs
```diff
@@ -81,10 +81,91 @@ mod chain_sync_test {
...
 
-        nigiri::try_cmd_repeatedly(nigiri::mine_blocks, 1, 10, FIFTH_SEC).unwrap();
+        nigiri::try_cmd_repeatedly(nigiri::mine_blocks, 1, 10, HALF_SEC).unwrap();
+
+        sleep(Duration::from_secs(10));
+
+        assert_eq!(node.get_node_info().num_channels, 0);
+    }
+
+    fn start_node_open_confirm_channel_stop_node(node_handle: &NodeHandle) -> String {
```
gcomte, line 91: Change the order to have all test cases first and helper functions afterwards

tests/chain_sync_test.rs
```diff
@@ -81,10 +81,91 @@ mod chain_sync_test {
...
+
+        sleep(Duration::from_secs(10));
+
+        assert_eq!(node.get_node_info().num_channels, 1);
+        assert_eq!(node.get_node_info().num_usable_channels, 1);
+
+        tx_id
+    }
+
+    fn start_node_open_channel_without_confirm_stop_node(node_handle: &NodeHandle) -> String {
```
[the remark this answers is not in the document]
danielgranhao: At the end of the function because `node` is dropped.

tests/chain_sync_test.rs
```diff
@@ -81,10 +81,91 @@ mod chain_sync_test {
...
+        //      as then we'll be able to handle not being connected to our peers
+        // nigiri::lnd_stop().unwrap();
+
+        nigiri::try_cmd_repeatedly(nigiri::mine_blocks, 1, 10, HALF_SEC).unwrap();
+
+        let node = node_handle.start().unwrap();
 
         sleep(Duration::from_secs(10));
 
+        // This only passes with the sleep that precedes it. TODO: confirm that's not a problem
```
gcomte, line 168: Is this comment really only relevant for this test case? The other's don't have that comment. Is it still required? Is the TODO still open?

src/storage_persister.rs
```diff
@@ -116,6 +119,18 @@ impl StoragePersister {
...
+                "Successfully read the ChannelManager from storage. It knows of {} channels",
+                channel_manager.list_channels().len()
+            );
+            debug!(
+                "List of channels known to the read ChannelManager: {}",
+                channel_manager
+                    .list_channels()
+                    .iter()
+                    .map(|details| details.channel_id.to_hex())
+                    .collect::<String>()
```
gcomte, line 132: Looks like multiple channel_id's won't be properly seperated

tests/chain_sync_test.rs
danielgranhao: It's mainly relevant here. In other situations is normal that we need to wait > 5 secs because we sync every 5 seconds. We wait 10 secs to be sure the node syncs. In this case, I'm not sure why the wait is needed because we sync on start.

tests/chain_sync_test.rs
```diff
@@ -81,10 +81,91 @@ mod chain_sync_test {
...
+    }
+
+    #[test]
+    fn test_channel_is_confirmed_chain_only_after_6_confirmations_offline_node() {
+        let node_handle = setup();
+
+        start_node_open_channel_without_confirm_stop_node(&node_handle);
+
+        nigiri::try_cmd_repeatedly(nigiri::mine_blocks, 6, 10, HALF_SEC).unwrap();
+        sleep(Duration::from_secs(5));
```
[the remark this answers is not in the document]
danielgranhao: The 10 seconds is for our node to sync. It's 10 to make sure sync happens in the meantime (it's every 5 secs). In here I found that a wait is needed, but 5 seconds is enough. I'm not sure exactly why it's needed, and I added now a TODO about that. I think it's related to either esplora or LND syncing.

gcomte (approved):

PR #83, opened 2022-10-26
Add our result

andrei-21:
- Add our own `Result` with our `Error`
- Make `lift_invalid_input()` a `Result` method
- Add `prefix_error()` handy method for `Result`
- Use the new error model for `encryption.rs`

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -58,15 +58,15 @@ pub fn permanent_failure_with<M: ToString + 'static, E: ToString>(
     })
 }
 
-pub type Result<T> = std::result::Result<T, LipaError>;
+pub type LipaResult<T> = Result<T, LipaError>;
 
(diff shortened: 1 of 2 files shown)


src/errors.rs
```diff
@@ -16,48 +16,77 @@ pub enum Error {
...
     })
 }
 
-pub fn lift_invalid_input(e: Error) -> Error {
-    match e {
-        Error::InvalidInput { message } => Error::PermanentFailure {
-            message: "InvalidInput: ".to_string() + &message,
-        },
-        another_error => another_error,
+pub type Result<T> = std::result::Result<T, Error>;
```
danielgranhao: I would rename this to `LipaResult` and the trait to another name. I think it can be confusing that the `Result` we use is not the std one.

danielgranhao (approved):

PR #88, opened 2022-10-26
Stabilize integration tests (when run locally)

gcomte:


tests/setup/mod.rs
```diff
@@ -99,17 +99,17 @@ pub mod nigiri {
...
-            debug!("NIGIRI stopping ...");
-            stop();
-        }
+            // TODO: Optimization, do not restart nigiri if it is already running in --ci mode:
+            //       `jq -r .ci  ~/.nigiri/nigiri.config.json` == true.
+            if env::var("RUNNING_ON_CI").is_err() {
+                debug!("NIGIRI stopping ...");
+                stop();
+            }
+        });
```
danielgranhao, line 109: This makes tests run faster, but it means that nigiri's state is kept between different integration tests. I think it's best if each integration test is independent of the others, so I would keep stopping nigiri before every INT test.

tests/setup/mod.rs
```diff
@@ -275,6 +273,26 @@ pub mod nigiri {
...
+    }
+
+    fn nigiri_runs_in_ci_mode() -> bool {
+        exec(vec!["jq", "-r", ".ci", &get_nigiri_config_path()])
+            .stdout
+            .starts_with(b"true")
+    }
+
+    fn get_nigiri_config_path() -> String {
+        env::var("HOME").unwrap() + "/.nigiri/nigiri.config.json"
```
danielgranhao, line 277: This assumes the default nigiri path for Linux. On macOS, the default is different -> `$HOME/Library/Application\ Support/Nigiri`

tests/setup/mod.rs
gcomte: Please note that this has been done locally only anyways: `if env::var("RUNNING_ON_CI").is_err()`. I perceived this more as a measure to ensure Nigiri is not running in a non-ci mode at the point in time when test are being started (which isn't the case when tests are run on CI anyways).

It is correct that Nigiri's state is being kept during all tests for now (and even before all of my changes, but only on CI, while my change produces the same behaviour locally as on CI), but the Lipa-node keeps forgetting it's old channels because its storage gets resetted between every test. That's why the tests all end up being successful.

I'll go ahead and add the additional functionality of resetting Nigiri after every test run.

gcomte: Changed here: 5ea7f513005ffef4fb557f364aeae2d80e5f0fda

tests/setup/mod.rs
gcomte: Becomes obsolete with: 5ea7f513005ffef4fb557f364aeae2d80e5f0fda

danielgranhao (approved):

PR #91, opened 2022-10-27
Integration test: Multiple txs at the same time

gcomte:
Add a chaotic test, that opens and closes multiple channels at the same time, to see whether the internal sorting of transactions and outputs works.

### DIFF
--- a/tests/chain_sync_test.rs
+++ b/tests/chain_sync_test.rs
@@ -145,7 +145,7 @@ mod chain_sync_test {
 
     fn setup() -> NodeHandle {
         nigiri::start();
-        let lsp_info = setup::nigiri::query_lnd_info().unwrap();
+        let lsp_info = nigiri::query_lnd_info().unwrap();
         let lsp_node = NodeAddress {
             pub_key: lsp_info.pub_key,
             address: "127.0.0.1:9735".to_string(),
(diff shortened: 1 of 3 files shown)


tests/chain_sync_test.rs
```diff
@@ -143,17 +143,40 @@ mod chain_sync_test {
...
+
+        // mine a block and do the same again and remove 1 of the previously opened channels
+        nigiri::try_cmd_repeatedly(nigiri::mine_blocks, 1, 10, HALF_SEC).unwrap();
+        let _ = open_5_chans_close_2(&node_id);
+        nigiri::lnd_force_close_channel(open_channels.pop().unwrap()).unwrap();
+
+        nigiri::try_cmd_repeatedly(nigiri::mine_blocks, 10, 10, HALF_SEC).unwrap();
+        sleep(Duration::from_secs(10));
+
+        assert_eq!(node.get_node_info().num_channels, 3);
```
danielgranhao, line 163: You can also check here for the number of usable channels. It should be 3 as well.

danielgranhao (approved):

tests/chain_sync_test.rs
gcomte: Yes, thanks for that great input. Fixed here: d0a7e80099ed9a2e53c1ef793612bd902857d03f

PR #92, opened 2022-11-03
Implement `mnemonic_to_secret`

danielgranhao:


src/keys_manager.rs
```diff
@@ -64,4 +92,14 @@ mod test {
         assert_eq!(secret.passphrase, "hodl");
         assert_eq!(secret.seed.len(), 32);
     }
+
+    #[test]
+    fn test_mnemonic_to_secret() {
```
gcomte, line 97: I would like to see a test where we actually test against hard-coded values. So generate a seed on https://iancoleman.io/bip39/ (use a password), and also copy the BIP39 seed from there and check whether we end up with the same result

src/keys_manager.rs
```diff
@@ -38,27 +38,32 @@ pub fn generate_secret(passphrase: String) -> Result<Secret, InitializationError
...
 
+    derive_secret_from_mnemonic(mnemonic, mnemonic_string, passphrase)
+}
+
+#[allow(clippy::result_large_err)]
+fn derive_secret_from_mnemonic(
+    mnemonic: Mnemonic,
+    mnemonic_string: Vec<String>,
+    passphrase: String,
+) -> Result<Secret, InitializationError> {
```
danielgranhao, line 66: Here we can just return a `Secret`

gcomte (approved):

PR #93, opened 2022-11-04
Reconnect to LSP node regularly in the background

gcomte:
Check whether the Lightning protocol connection to the LSP is up every second, and if not try to reconnect.

In the current form, the reconnecting only happens towards the LSP node, which will also be the only Lightning peer for the foreseeable future.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -38,7 +38,7 @@ use crate::keys_manager::{
 };
 use crate::logger::LightningLogger;
 use crate::native_logger::init_native_logger_once;
-use crate::p2p_networking::{LnPeer, P2pConnections};
+use crate::p2p_networking::{LnPeer, P2pConnection};
 use crate::secret::Secret;
 use crate::storage_persister::StoragePersister;
 use crate::tx_broadcaster::TxBroadcaster;
@@ -70,7 +70,7 @@ pub struct LightningNode {
     background_processor: BackgroundProcessor,
     channel_manager: Arc<ChannelManager>,
     peer_manager: Arc<PeerManager>,
-    peer_conn_handle: JoinHandle<()>,
+    p2p_connector_handle: JoinHandle<()>,
     sync_handle: JoinHandle<()>,
 }
 
@@ -191,8 +191,9 @@ impl LightningNode {
         )?);
 
         // Step 13. Initialize Networking
-        let peer_conn_handle =
-            P2pConnections::run_bg_connector(&config.lsp_node, rt.handle(), &peer_manager).unwrap(); // todo proper error handling instead of unwrap()
+        let p2p_connector_handle =
+            P2pConnection::init_background_task(&config.lsp_node, rt.handle(), &peer_manager)
+                .unwrap(); // todo proper error handling instead of unwrap()
 
         // Step 14. Keep LDK Up-to-date with Chain Info
         // TODO: optimize how often we want to run sync. LDK-sample syncs every second and
@@ -266,7 +267,7 @@ impl LightningNode {
             background_processor,
             channel_manager: Arc::clone(&channel_manager),
             peer_manager,
-            peer_conn_handle,
+            p2p_connector_handle,
             sync_handle,
         })
     }
@@ -293,7 +294,7 @@ impl LightningNode {
 
 impl Drop for LightningNode {
     fn drop(&mut self) {
-        self.peer_conn_handle.abort();
+        self.p2p_connector_handle.abort();
         self.sync_handle.abort();
 
         // TODO: Stop reconnecting to peers
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -204,7 +204,7 @@ impl LightningNode {
 
         // Step 14. Keep LDK Up-to-date with Chain Info
         // TODO: optimize how often we want to run sync. LDK-sample syncs every second and
-        // LDKLite syncs every 5 seconds. Let's try 5 seconds first and change if needed
+        //       LDKLite syncs every 5 seconds. Let's try 5 seconds first and change if needed
         let channel_manager_regular_sync = Arc::clone(&channel_manager);
         let chain_monitor_regular_sync = Arc::clone(&chain_monitor);
         let sync_handle = rt
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -38,7 +38,7 @@ use crate::keys_manager::{
 };
 use crate::logger::LightningLogger;
 use crate::native_logger::init_native_logger_once;
-use crate::p2p_networking::P2pConnections;
+use crate::p2p_networking::{LnPeer, P2pConnections};
 use crate::secret::Secret;
 use crate::storage_persister::StoragePersister;
 use crate::tx_broadcaster::TxBroadcaster;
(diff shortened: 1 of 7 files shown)


src/p2p_networking.rs
```diff
@@ -1,25 +1,50 @@
...
+use tokio::task::JoinHandle;
 use tokio::time::sleep;
 
 pub(crate) struct P2pConnections {}
 
 impl P2pConnections {
-    pub async fn connect_peer(
+    pub fn run_bg_connector(
         peer: &NodeAddress,
+        handle1: Handle,
```
danielgranhao, line 20: Maybe rename it to `runtime_handle` or `tokio_handle`?

src/p2p_networking.rs
```diff
@@ -1,25 +1,50 @@
...
+        peer_mgr: &Arc<PeerManager>,
```
danielgranhao, line 21: To maintain consistency and readability, I would prefer we use `peer_manager`

danielgranhao: Oh, I see that maybe you used this variation to distinguish from the inner `peer_manager`? I would still avoid doing so because we can use shadowing.

src/p2p_networking.rs
```diff
@@ -1,25 +1,50 @@
...
 use std::task::Poll;
 use std::time::Duration;
+use tokio::task::JoinHandle;
 use tokio::time::sleep;
 
 pub(crate) struct P2pConnections {}
 
 impl P2pConnections {
-    pub async fn connect_peer(
+    pub fn run_bg_connector(
```
danielgranhao, line 18: I would prefer `start_p2p_connector` or `start_p2p_background_connector`

src/lib.rs
```diff
@@ -70,6 +70,7 @@ pub struct LightningNode {
     background_processor: BackgroundProcessor,
     channel_manager: Arc<ChannelManager>,
     peer_manager: Arc<PeerManager>,
+    peer_conn_handle: JoinHandle<()>,
```
danielgranhao, line 73: I would rename it to `p2p_connector_handle` otherwise, it seems like it is a handle to a connection and not to the connector task.

tests/p2p_connection_test.rs
```diff
@@ -6,46 +6,124 @@ mod setup;
...
+        };
+
+        // Let's shutdown LND, while we leave Esplora running (test should not affect chain sync)
+        setup::nigiri::pause();
+        setup::nigiri::resume_without_ln();
+
+        let node = NodeHandle::new(lsp_node.clone()).start().unwrap();
+
+        // Wait for the LDK to connect to the LSP (which should fail)
+        sleep(Duration::from_millis(1500));
```
danielgranhao, line 63: We can remove the `sleep()` as it's already being done inside `start()` right?

danielgranhao: The other option is to remove the sleep from `start()`. Maybe this latter option is best because we'll need to sleep anyway after restarting nigiri and it may get confusing when a sleep is needed and when it's not.

tests/p2p_connection_test.rs
```diff
@@ -6,46 +6,124 @@ mod setup;
...
+        setup::nigiri::start();
+        let lsp_info = setup::nigiri::query_lnd_info().unwrap();
+        let lsp_node = NodeAddress {
+            pub_key: lsp_info.pub_key,
+            address: NIGIRI_LND_ADDR.to_string(),
+        };
+
+        {
+            let node = NodeHandle::new(lsp_node.clone()).start().unwrap();
+
```
danielgranhao, line 90: In case we remove the sleep from `start()`, we need to add it here.

src/p2p_networking.rs
gcomte: Oh yeah, that was meant to be changed before commiting it :grimacing:

src/p2p_networking.rs
```diff
@@ -68,26 +93,50 @@ impl P2pConnections {
...
-                message: e.to_string(),
-            }
-        })?;
-        let address = SocketAddr::from_str(&node_address.address).map_err(|e| {
-            RuntimeError::InvalidAddress {
-                message: e.to_string(),
-            }
-        })?;
+        let pub_key = PublicKey::from_str(&node_address.pub_key)
+            .map_to_invalid_input("Could not parse node public key")?;
```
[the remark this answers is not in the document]
gcomte: Fixed: be587ed4cc555048b652e265daa899f51f637e2d

src/p2p_networking.rs
gcomte: Fixed: 440c22abbf62ab577edd2f83b871a1af427ca4ef

src/p2p_networking.rs
gcomte: Fixed: 440c22abbf62ab577edd2f83b871a1af427ca4ef

src/p2p_networking.rs
gcomte: how about `init` instead of `start`? I took out the `p2p` part, because it is a function of the struct `P2pConnections`, so whenever you call it you'd run P2pConnections::start_p2p_background_connector which is somewhat redundant.

In fact I'd kind of propose the opposite, and removing both connection and p2p because of the redundancy and instead call it something like:
`P2pConnections::init_background_service()` or `P2pConnections::init_background_task()`.

Also, imo `P2pConnections` should be renamed to `P2pConnection`, because so far it is really only dealing with one connection, namely the connection with the LSP.

WDYT?

danielgranhao: > how about init instead of start? I took out the p2p part, because it is a function of the struct P2pConnections, so whenever you call it you'd run P2pConnections::start_p2p_background_connector which is somewhat redundant.

> In fact I'd kind of propose the opposite, and removing both connection and p2p because of the redundancy and instead call it something like:
> P2pConnections::init_background_service() or P2pConnections::init_background_task().

Agreed. `P2pConnections::init_background_service()` or `P2pConnections::init_background_task()` SGTM. 

> Also, imo P2pConnections should be renamed to P2pConnection, because so far it is really only dealing with one connection, namely the connection with the LSP.

I think we might quickly want to have more than one LSP node but for now yes, we can change to singular.

tests/p2p_connection_test.rs
gcomte: The way I understand it, is that the `start()` function waits until Nigiri is up and running. However, that does not necessarily mean that LDK already reconnected to the now started LND instance. Remember that it retries every second.

That given, I think both the `sleep` within `start()` and the `sleep` being discussed here have their own importance and should not be removed.

src/p2p_networking.rs
gcomte: Fixed: aa2f6c6ad837b014a771429f51d9cc5fa2db57de

src/lib.rs
gcomte: Fixed: aa2f6c6ad837b014a771429f51d9cc5fa2db57de

tests/p2p_connection_test.rs
gcomte: Obsolete (see discussion above)

danielgranhao (approved):

PR #95, opened 2022-11-07
Make `lspd` container wait for `db` container

andrei-21:
Now `docker-compose up lspd` does not fail because of starting `db`.

### DIFF
--- a/lspd/docker-compose.yml
+++ b/lspd/docker-compose.yml
@@ -9,6 +9,12 @@ services:
       POSTGRES_USER: lipauser
       POSTGRES_PASSWORD: lipapassword
       POSTGRES_DB: lipadb
+    healthcheck:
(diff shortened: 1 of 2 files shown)


lspd/docker-compose.yml
```diff
@@ -9,6 +9,12 @@ services:
       POSTGRES_USER: lipauser
       POSTGRES_PASSWORD: lipapassword
       POSTGRES_DB: lipadb
+    healthcheck:
+      test: ["CMD", "psql", "-U", "lipauser", "lipadb"]
```
danielgranhao, line 13: I think this can be made simpler and more generalised.
```suggestion
      test: ["CMD-SHELL", "pg_isready"]
```

danielgranhao (approved):
I had already worked on something equivalent as part of #73 but we can go ahead with this PR instead. I just propose a small improvement.

gcomte (approved):

PR #101, opened 2022-11-09
Add routing hints for unannouced channels

andrei-21:


tests/zero_conf_test.rs
```diff
@@ -22,10 +22,25 @@ mod zero_conf_test {
...
 
+        const TEN_SATS: u64 = 10_000;
+
+        // With no channels, 10 sats invoice is to small to cover channel
+        // opening fees.
+        let invoice = node.create_invoice(TEN_SATS, "test".to_string());
+        assert_eq!(
+	    invoice.unwrap_err().to_string(),
+	    "PermanentFailure: Failed to register payment: InvalidInput: Payment amount must be bigger than fees");
+        let invoice = node.create_invoice(20_000_000, "test".to_string());
```
gcomte, line 33: `const TWENTY_K_SATS = 20_000_000` ?

Or why using a constant for one of the numbers but not the other one?

gcomte (commented):
I don't grasp the relationship between `channel capacity`, `channel reserve`, and `inbound capacity`.

Do we need a test to clarify, resp. to make sure that the we don't allow for payments that would empty our counter party's funds beyond their channel reserve?

danielgranhao (approved):

gcomte (approved):

PR #111, opened 2022-11-17
Simplify sync

andrei-21:
The new implementation is simpler and cleaner in my humble opinion.
Although it is not optimal:
* in case if a new block is mined on top of an existent one, sync attempt is aborted (since it is not feasible to prove that there were no reorg)
* if during the sync new outputs to watch were got the implementation will recheck **all** transactions
* the implementation keeps **all** outputs watching, event spent ones (it is due to LDK limitation, to be clarified with LDK devs)

### DIFF
--- a/src/chain_access.rs
+++ b/src/chain_access.rs
@@ -4,7 +4,7 @@ use crate::filter::FilterImpl;
 use crate::ConfirmWrapper;
 use crate::EsploraClient;
 
-use bitcoin::{BlockHash, BlockHeader, Txid};
+use bitcoin::{BlockHash, Txid};
 use lightning::chain::{Confirm, WatchedOutput};
 use log::debug;
 use std::collections::HashSet;
@@ -18,15 +18,9 @@ pub(crate) struct LipaChainAccess {
     synced_tip: BlockHash,
 }
 
-struct BlockInfo {
-    pub hash: BlockHash,
-    pub header: BlockHeader,
-    pub height: u32,
-}
-
 enum SyncResult {
     Success,
-    PossibleReorg,
+    PotentialReorg,
 }
 
 impl LipaChainAccess {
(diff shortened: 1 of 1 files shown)


danielgranhao (approved):

src/chain_access.rs
```diff
@@ -35,146 +45,156 @@ impl LipaChainAccess {
...
-        //       otherwise, we might end up in a situation where a block is being mined while while we're syncing
-        //       which would make corrupt the order of the confirmed transactions as required by the Confirm trait:
-        //       https://github.com/lightningdevkit/rust-lightning/blob/d6321e6e11b3e1b11e26617d3bab0b8c21da0b5b/lightning/src/chain/mod.rs#L128
+    fn try_sync_to(
+        &mut self,
+        tip_hash: BlockHash,
+        confirm: &ConfirmWrapper,
+    ) -> LipaResult<SyncResult> {
+        let tip;
+        if let Some(tip_val) = self.get_block_info(tip_hash)? {
```
gcomte, line 84: An `if let` statement that also has an `else` clause => Looks to me like a `match` statement would be a better fit

src/chain_access.rs
```diff
@@ -35,146 +45,156 @@ impl LipaChainAccess {
...
-                continue;
+    fn split_txs_by_status(
+        &self,
+        txids: HashSet<Txid>,
+    ) -> LipaResult<(Vec<ConfirmedTransaction>, Vec<Txid>)> {
+        let mut confirmed = Vec::new();
+        let mut pending = Vec::new();
+
+        for txid in txids {
+            if let Some(tx) = self.esplora.get_confirmed_tx_by_id(&txid)? {
```
gcomte, line 159: Here again, imo an `if let` statement that also has an `else` clause is better off being replaced by a `match` statement.

gcomte (commented):
The implementation is much more clean now! 👏

gcomte (approved):

PR #115, opened 2022-11-18
Test RGS

danielgranhao:
This is building on top of #112 and #114. Leaving as a draft until they are merged.

Currently, it's only checking that an update could be obtained from RGS and applied to the local network graph. More comprehensive testing will be possible when we implement sending payments.

### DIFF
--- a/tests/rapid_gossip_sync_test.rs
+++ b/tests/rapid_gossip_sync_test.rs
@@ -1,0 +1,92 @@
+mod setup;
+
+// Caution: Run these tests sequentially, otherwise they will corrupt each other
+//      because they are manipulating their environment:
+// cargo test --features nigiri -- --test-threads 1
+#[cfg(feature = "nigiri")]
+mod zero_conf_test {
+    use crate::setup::nigiri::NodeInstance;
+    use crate::setup::{nigiri, NodeHandle};
+    use crate::try_cmd_repeatedly;
+    use bitcoin::hashes::hex::ToHex;
+    use std::thread::sleep;
+    use std::time::Duration;
+
+    const TWENTY_K_SATS: u64 = 20_000_000;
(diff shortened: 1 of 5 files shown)


tests/rapid_gossip_sync_test.rs
```diff
@@ -0,0 +1,101 @@
+mod setup;
+
+// Caution: Run these tests sequentially, otherwise they will corrupt each other
+//      because they are manipulating their environment:
+// cargo test --features nigiri -- --test-threads 1
```
gcomte, line 5:
```suggestion
// Caution: Run these tests sequentially, otherwise they will corrupt each other
//          because they are manipulating their environment:
//          cargo test --features nigiri -- --test-threads 1
```

tests/setup/mod.rs
```diff
@@ -163,16 +186,21 @@ pub mod nigiri {
 
         start_nigiri();
         start_lspd();
+        start_rgs();
     }
 
     pub fn stop() {
+        debug!("RGS server stopping ...");
```
gcomte, line 193: would it make sense to move the debug log to the `stop_rgs` function, to keep it consistent with the `start_rgs` function?

tests/setup/mod.rs
```diff
@@ -163,16 +186,21 @@ pub mod nigiri {
...
+        stop_rgs();
         debug!("LSPD stopping ...");
         stop_lspd(); // Nigiri cannot be stopped if lspd is still connected to it.
         debug!("NIGIRI stopping ...");
         exec(&["nigiri", "stop", "--delete"]);
     }
 
     pub fn pause() {
+        debug!("RGS server stopping ...");
```
gcomte, line 202: would it make sense to move the debug log to the `stop_rgs` function, to keep it consistent with the `start_rgs` function?

gcomte (approved):

PR #124, opened 2022-12-15
Use `RemoteStorageCallback`

andrei-21:
Almost blindly replace `RedundatStorageCallback` with `RemoteStorageCallback`, what does not make much sense.
I will create stories to for a proper implementation.

### DIFF
--- a/src/storage_persister.rs
+++ b/src/storage_persister.rs
@@ -105,7 +105,7 @@ impl StoragePersister {
         if self
             .storage
             .object_exists(OBJECTS_BUCKET.to_string(), MANAGER_KEY.to_string())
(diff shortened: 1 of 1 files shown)


src/storage_persister.rs
```diff
@@ -102,10 +105,12 @@ impl StoragePersister {
         if self
             .storage
             .object_exists(OBJECTS_BUCKET.to_string(), MANAGER_KEY.to_string())
+            .map_to_runtime_error("Failed to check read channel manager")?
```
danielgranhao, line 108:
```suggestion
            .map_to_runtime_error("Failed to check channel manager")?
```

danielgranhao (approved):

gcomte (approved):

PR #125, opened 2022-12-15
Cancel in-progress actions

danielgranhao:
If we push a lot of commits sequentially, the CI jobs get queued and take a long time. This should make it so that adding a new commit cancels in-progress jobs for the same branch.


gcomte:
So cool, didn't know that was possible

gcomte (approved):

PR #154, opened 2023-01-04
Learn LSP LN node from LSP itself

andrei-21:
The idea here is to introduce `TaskManager` responsible for all background tasks.
The first part is two task: one is it to query LSP info, second to connect to LSP LN node. It required some refactoring because before we were requiring to configure it for the node. But now the node learns LN address itself from LSP and connects to it.

### DIFF
--- a/docs/LifetimeModel.org
+++ b/docs/LifetimeModel.org
@@ -15,30 +15,26 @@
 
 * Background Tasks
 1. check remote channel monitors if they are newer
-2. upload channel monitors to remote if the previous attempt failed???
-3. ~lightning_background_processor::BackgroundProcessor~
+2. ~lightning_background_processor::BackgroundProcessor~
    1. process ~Event~s
    2. persist ~ChannelManager~
-4. sync blockchain
-5. query on-chain fees
-6. update network graph
-7. query LSP info
-8. reconnect to LSP LN node
-9. reconnect to channel peers
+3. sync blockchain
+4. query on-chain fees
+5. update network graph
+6. query LSP info
+7. reconnect to LSP LN node
+8. reconnect to channel peers
 
 * Lifetime Model
 |---+----------------------------+---------+------------+------------|
 |   | Task                       | Startup | Foreground | Background |
 |---+----------------------------+---------+------------+------------|
 | 1 | check remote *monitors*    | yes     | as sync    | as sync    |
-| 2 | upload *monitors*          | wait    | 60 secs    |            |
-| 3 | run *BackgroundProcessor*  | wait    | yes        | yes?       |
-| 4 | sync *blockchain*          | wait    | 5 mins     | 1 hour     |
-| 5 | query *on-chain fees*      | yes     | 5 mins     |            |
-| 6 | update *network graph*     | yes     | 1 hour     |            |
-| 7 | query *lsp info*           | yes     | 10 mins    |            |
-| 8 | reconnect to *LSP LN node* | yes     | 10 sec     | 1 min      |
-| 9 | reconnect to *peers*       | yes     | 10 secs    | 1 min      |
+| 2 | run *BackgroundProcessor*  | wait    | yes        | yes?       |
+| 3 | sync *blockchain*          | wait    | 5 mins     | 1 hour     |
+| 4 | query *on-chain fees*      | yes     | 5 mins     |            |
+| 5 | update *network graph*     | yes     | 1 hour     |            |
+| 6 | query *lsp info*           | yes     | 10 mins    |            |
+| 7 | reconnect to *LSP LN node* | yes     | 10 sec     | 1 min      |
+| 8 | reconnect to *peers*       | yes     | 10 secs    | 1 min      |
 |---+----------------------------+---------+------------+------------|
-
-** Shutdown
--- a/src/task_manager.rs
+++ b/src/task_manager.rs
@@ -41,14 +41,14 @@ impl TaskManager {
         (*self.lsp_info.lock().unwrap()).clone()
     }
 
-    pub fn request_shutdowns(&mut self) {
+    pub fn request_shutdown_all(&mut self) {
         self.task_handles
             .drain(..)
             .for_each(|h| h.request_shutdown());
     }
 
     pub fn restart(&mut self, periods: TaskPeriods) {
-        self.request_shutdowns();
+        self.request_shutdown_all();
 
         // TODO: Blockchain sync.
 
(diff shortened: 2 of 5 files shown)


docs/LifetimeModel.org
```diff
@@ -0,0 +1,44 @@
...
+* Background Tasks
+1. check remote channel monitors if they are newer
+2. upload channel monitors to remote if the previous attempt failed???
+3. ~lightning_background_processor::BackgroundProcessor~
+   1. process ~Event~s
+   2. persist ~ChannelManager~
+4. sync blockchain
+5. query on-chain fees
+6. update network graph
+7. query LPS info
```
gcomte, line 25:
```suggestion
7. query LSP info
```

docs/LifetimeModel.org
```diff
@@ -0,0 +1,44 @@
...
+* Lifetime Model
+|---+----------------------------+---------+------------+------------|
+|   | Task                       | Startup | Foreground | Background |
+|---+----------------------------+---------+------------+------------|
+| 1 | check remote *monitors*    | yes     | as sync    | as sync    |
+| 2 | upload *monitors*          | wait    | 60 secs    |            |
+| 3 | run *BackgroundProcessor*  | wait    | yes        | yes?       |
+| 4 | sync *blockchain*          | wait    | 5 mins     | 1 hour     |
+| 5 | query *on-chain fees*      | yes     | 5 mins     |            |
+| 6 | update *network graph*     | yes     | 1 hour     |            |
```
gcomte, line 38: Remember that the RGS only updates itself every 24 hours  if I'm not mistaken ( @danielgranhao )

danielgranhao: Yep, that's true!

src/task_manager.rs
```diff
@@ -0,0 +1,128 @@
...
+        let lsp_info = Arc::clone(&self.lsp_info);
+        self.runtime_handle.spawn_repeating_task(period, move || {
+            let peer_manager = Arc::clone(&peer_manager);
+            let lsp_client = Arc::clone(&lsp_client);
+            let lsp_info = Arc::clone(&lsp_info);
+            async move {
+                let result = tokio::task::spawn_blocking(move || lsp_client.query_info()).await;
+                match result {
+                    Ok(Ok(new_lsp_info)) => {
+                        debug!("New LSP info: {:?}", new_lsp_info);
```
danielgranhao, line 83: This could be confusing if the LSP info didn't change. We could print this if the LSP info differs from the one we know. Otherwise, print "Queried LSP for changes... No changes to LSP info." or something similar.

tests/p2p_connection_test.rs
```diff
@@ -15,119 +15,45 @@ mod p2p_connection_test {
...
+            pub_key: lsp_info.pub_key,
+            host: NIGIRI_LND_ADDR.to_string(),
+        };
 
-            let node = NodeHandle::new(lsp_node.clone()).start().unwrap();
+        let node = NodeHandle::new(lsp_node.clone()).start().unwrap();
 
+        // Test successful p2p connection.
+        {
+            sleep(Duration::from_secs(100));
```
danielgranhao, line 33: We should try avoid these long periods of sleep whenever possible. Would you happen to know why it's needed here? If it's because of the `TaskPeriods` we could have a test configuration that makes tasks run more frequently.

tests/p2p_connection_test.rs
```diff
@@ -15,119 +15,45 @@ mod p2p_connection_test {
...
-
-            // Let's shutdown LND, while we leave Esplora running (test should not affect chain sync)
-            setup::nigiri::pause();
-            setup::nigiri::resume_without_ln();
-
-            let node = NodeHandle::new(lsp_node.clone()).start().unwrap();
-
-            assert_eq!(node.get_node_info().num_peers, 0);
-
             // Now let's start Nigiri with LND again
```
danielgranhao, line 50:
```suggestion
            // Now let's start LSPD-LND again
```

tests/p2p_connection_test.rs
```diff
@@ -15,119 +15,45 @@ mod p2p_connection_test {
...
-            let lsp_info = setup::nigiri::query_node_info(NodeInstance::NigiriLnd).unwrap();
-            let lsp_node = NodeAddress {
-                pub_key: lsp_info.pub_key,
-                host: NIGIRI_LND_ADDR.to_string(),
-            };
-            let node = NodeHandle::new(lsp_node.clone()).start().unwrap();
-
+            setup::nigiri::start_lspd();
+            setup::nigiri::wait_for_healthy_lspd();
+            sleep(Duration::from_secs(10));
```
danielgranhao, line 53: Same as in the other comment. If we run the reconnect task more frequently in tests (e.g. 1 sec), we can decrease this sleep to 1-2 secs.

docs/LifetimeModel.org
```diff
@@ -0,0 +1,44 @@
...
+| 7 | query *lsp info*           | yes     | 10 mins    |            |
+| 8 | reconnect to *LSP LN node* | yes     | 10 sec     | 1 min      |
+| 9 | reconnect to *peers*       | yes     | 10 secs    | 1 min      |
```
danielgranhao, line 41: It makes sense to run these reconnection tasks more frequently than every 10 secs. Maybe 1/2 secs? 10 secs means that if the phone loses network connection for a brief moment when it gets back online, the user still has to wait 10 more seconds.

docs/LifetimeModel.org
```diff
@@ -0,0 +1,44 @@
...
+
+* Lifetime Model
+|---+----------------------------+---------+------------+------------|
+|   | Task                       | Startup | Foreground | Background |
+|---+----------------------------+---------+------------+------------|
+| 1 | check remote *monitors*    | yes     | as sync    | as sync    |
+| 2 | upload *monitors*          | wait    | 60 secs    |            |
+| 3 | run *BackgroundProcessor*  | wait    | yes        | yes?       |
+| 4 | sync *blockchain*          | wait    | 5 mins     | 1 hour     |
+| 5 | query *on-chain fees*      | yes     | 5 mins     |            |
```
danielgranhao, line 37: Is a 5 min interval frequent enough? I feel like fees can change a lot in 5 mins, but maybe that's not critical for us. Something to think about.

docs/LifetimeModel.org
```diff
@@ -0,0 +1,44 @@
...
+7. query LPS info
+8. reconnect to LSP LN node
+9. reconnect to channel peers
+
+* Lifetime Model
+|---+----------------------------+---------+------------+------------|
+|   | Task                       | Startup | Foreground | Background |
+|---+----------------------------+---------+------------+------------|
+| 1 | check remote *monitors*    | yes     | as sync    | as sync    |
+| 2 | upload *monitors*          | wait    | 60 secs    |            |
```
danielgranhao, line 34: I don't think we can update the monitors on a periodic basis. We must guarantee that channel states in the phone and in the remote service are always in sync. This means that we push new monitors every time the monitors change.

docs/LifetimeModel.org
danielgranhao: The likelihood of the node running for 1 hour is very low, so I would even say to do it on startup or on coming into the foreground. Trying to update the network graph more than necessary is also likely not to result in significant overhead in network traffic or battery usage due to the server only returning necessary data (i.e. nothing if we are synced).

danielgranhao (approved):

PR #160, opened 2023-01-05
Fix double task execution

andrei-21:
```rust
interval.tick().await;
// approximately 0ms have elapsed. The first tick completes immediately.
```

https://docs.rs/tokio/latest/tokio/time/struct.Interval.html#method.tick


gcomte (approved):
interesting!

danielgranhao (approved):

PR #168, opened 2023-01-11
Implement ChannelManager remote persistence

danielgranhao:


danielgranhao:
Dropped in favour of #169

PR #178, opened 2023-01-25
Update network graph

andrei-21:
As the config says we update the graph when the app goes foreground once if succeeds or retry in 2 minutes.

### DIFF
--- a/eel/src/task_manager.rs
+++ b/eel/src/task_manager.rs
@@ -80,6 +80,9 @@ impl TaskManager {
         }
 
         // Update network graph.
+        // We regularly retry to update the network graph if it fails.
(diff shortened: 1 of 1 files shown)


eel/src/task_manager.rs
```diff
@@ -71,7 +79,10 @@ impl TaskManager {
             self.task_handles.push(self.start_fee_update(period));
         }
 
-        // TODO: Update network graph.
+        // Update network graph.
```
danielgranhao, line 82:
```suggestion
        // Update network graph. We regularly retry to update the network graph if it fails. After the first successful update, not further updates are tried until the app gets to the foreground again.
```

danielgranhao (approved):

PR #201, opened 2023-02-02
Ensure symmetric key size by strong typing

gcomte:
Use strong typing to ensure the passed in key is 32 bytes long, instead of throwing an error.

### DIFF
--- a/eel/src/encryption_symmetric.rs
+++ b/eel/src/encryption_symmetric.rs
@@ -12,9 +12,7 @@ use perro::MapToError;
 type NonceLength = U12;
 type Nonce = AesNonce<NonceLength>;
 
-const KEY_SIZE_IN_BYTES: usize = 32;
-
-pub(crate) fn encrypt(data: &[u8], key: &[u8; KEY_SIZE_IN_BYTES]) -> Result<Vec<u8>> {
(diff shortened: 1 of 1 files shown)


eel/src/encryption_symmetric.rs
```diff
@@ -21,7 +23,7 @@ pub(crate) fn encrypt(data: &[u8], key: &[u8]) -> Result<Vec<u8>> {
     Ok(ciphertext)
 }
 
-pub(crate) fn decrypt(data: &[u8], key: &[u8]) -> Result<Vec<u8>> {
+pub(crate) fn decrypt(data: &[u8], key: &[u8; KEY_SIZE_IN_BYTES]) -> Result<Vec<u8>> {
```
danielgranhao, line 26: nit: In this specific case (length of `key`) I would get rid of the constant and use `32` directly. I feel like the constant doesn't add much given that we know that 32 is the key size as it's the array length of an array reference called `key`. A reason to not use it is that the code becomes more concise.

That is my personal preference, but feel free to leave it like this if you disagree.

danielgranhao (approved):

eel/src/encryption_symmetric.rs
gcomte: 6d9e196a8137532f4202f641af419108b601ca88

PR #202, opened 2023-02-02
Reject symmetric encryption with obviously bad key

gcomte:


gcomte:
Deemed as being checked in the wrong place.

PR #203, opened 2023-02-02
Encrypt remote storage

gcomte:


eel/src/config.rs
```diff
@@ -2,8 +2,16 @@ use bitcoin::Network;
...
     pub network: Network,
-    pub seed: Vec<u8>,
+    pub seed: [u8; 64],
     pub esplora_api_url: String,
     pub rgs_url: String,
     pub local_persistence_path: String,
 }
+
+impl Config {
+    pub fn get_first_half_of_seed(&self) -> [u8; 32] {
```
danielgranhao, line 12: nit on naming:
```suggestion
    pub fn get_seed_first_half(&self) -> [u8; 32] {
```

eel/src/keys_manager.rs
```diff
@@ -37,12 +32,12 @@ pub fn mnemonic_to_secret(mnemonic_string: Vec<String>, passphrase: String) -> R
 }
 
 fn derive_secret_from_mnemonic(mnemonic: Mnemonic, passphrase: String) -> Secret {
```
danielgranhao, line 34: I think it makes sense to move this logic (secret generation, mnemonic to secret, etc) from `eel` to `3L`. But let's do that in another PR.

eel/src/storage_persister.rs
```diff
@@ -447,13 +457,26 @@ impl<ChannelSigner: Sign> Persist<ChannelSigner> for StoragePersister {
...
     data: Vec<u8>,
     channel_id: OutPoint,
 ) {
     let key = channel_id.to_channel_id().to_hex();
+    let encrypted_data = match encrypt(&data, &encryption_key) {
+        Ok(data) => data,
+        Err(e) => {
+            error!("Failed to encrypt the ChannelMonitor: {}", e.to_string());
+            return;
+        }
```
danielgranhao, line 471: Here it makes sense to return an error and make `persist_new_channel()` return `ChannelMonitorUpdateStatus::PermanentFailure` in case `sync_persist_monitor_remotely` fails. I've already made this change on PR #204, so let's **not** do it here. When we rebase either this PR or the other, we fix it.

eel/src/storage_persister.rs
```diff
@@ -512,12 +535,22 @@ where
         };
 
         // Persist remotely
+        let encrypted_data = match encrypt(&channel_manager.encode(), &self.encryption_key) {
+            Ok(data) => data,
+            Err(e) => {
+                error!("Failed to encrypt the ChannelManager: {}", e.to_string());
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::Other,
+                    "Failed to encrypt the ChannelManager",
```
danielgranhao, line 544:
```suggestion
                    format!("Failed to encrypt the ChannelManager: {}", e.to_string()),
```

danielgranhao (commented):
You're only missing decrypting `ChannelMonitors` and the `ChannelManager` when they are obtained from remote, right?

eel/src/config.rs
gcomte: c19cba90d0f41b6825aae1039c11d6f18760a38c

eel/src/storage_persister.rs
gcomte: Applied, but then had to remove the `.to_string()`: a65fb331ee917f8409816426e04a75fe130f1c30

eel/src/storage_persister.rs
gcomte: I'm not sure if I understand you correctly. Rebasing onto #204 forced me to make a change here anyways, so could you please have a look at it again?

danielgranhao (approved):

eel/src/storage_persister.rs
danielgranhao: Forget that, you did exactly what I meant. Only fix after rebasing, not before :)

eel/src/keys_manager.rs
gcomte: I've created a Jira task.

PR #205, opened 2023-02-03
Refactor `read_channel_monitors()`

danielgranhao:


danielgranhao:
Dropped in favour of #206

PR #209, opened 2023-02-08
Implement csp client

gcomte:


.github/workflows/build.yml
```diff
@@ -35,6 +35,8 @@ jobs:
         uses: Swatinem/rust-cache@v2.2.0
       - name: Build
         run: cargo build --verbose
+      - name: Set environment variables
+        run: echo "BACKEND_BASE_URL=${{ secrets.BACKEND_BASE_URL }}" >> $GITHUB_ENV
       - name: Run unit tests
         run: make test
```
danielgranhao, line 41:
```suggestion
      - name: Run unit tests
        env:
          BACKEND_BASE_URL: ${{ secrets.BACKEND_BASE_URL }}
        run: make test
```

src/auth.rs
```diff
@@ -0,0 +1,80 @@
...
+// Since this code right here is using a sensible key derivation path from the get-go,
+// this scheme would not be necessary. However, the backend still expects the old scheme,
+// so that's why we just derive 2 key pairs, and follow the overhead of the old scheme.
+// For the ones who have access, follow this issue: https://getlipa.atlassian.net/browse/APP-1057
+const BACKEND_WALLET_DERIVATION_PATH: &str = "m/76738065'/0'/0";
+const BACKEND_AUTH_DERIVATION_PATH: &str = "m/76738065'/0'/2";
+
+pub struct Auth {
+    auth: Arc<honey_badger::Auth>,
+}
```
danielgranhao, line 25: Why do we need this new wrapper `Auth` struct? On `lib.rs`, we already have the `build_auth()` function which returns an instance of `honey_badger::Auth` and is used for the exchange rate provider.

danielgranhao: Between both approaches, I would go with the more compact `build_auth()` function.

tests/remote_persistence_test.rs
```diff
@@ -0,0 +1,69 @@
...
+    const HALF_SEC: Duration = Duration::from_millis(500);
+    const N_RETRIES: u8 = 10;
+
+    #[test]
+    #[file_serial(key, "/tmp/3l-int-tests-lock")]
+    fn test_remote_persistence() {
+        nigiri::setup_environment_with_lsp();
+
+        let node_handle = NodeHandle::new();
+
```
danielgranhao, line 29: We should think about how to avoid bloating the backend CSP service with test channel states. Maybe use a specific tag on the custom fields and use that to clean the db? Another option is to mock `mole` such that we still test the integration between `eel` and `mole` without needing the backend.

tests/remote_persistence_test.rs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            let node = node_handle.start().unwrap();
+            let node_info = node.get_node_info();
+            let node_id = node_info.node_pubkey.to_hex();
+            assert_eq!(node_info.num_peers, 1);
+
+            log::error!("before opening channels");
```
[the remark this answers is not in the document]
gcomte: Nope, that should defenitely not be there. thanks.

src/auth.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub fn get_instace(&self) -> Arc<honey_badger::Auth> {
+        Arc::clone(&self.auth)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::env;
+    use std::thread;
```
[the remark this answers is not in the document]
gcomte: thanks

src/auth.rs
gcomte: YES: fbc78398206eeceae1fd5e4903bf36e147f9a6a7

tests/remote_persistence_test.rs
gcomte: I don't necessarily agree that it is `3L`'s responsibility to keep the backend database clean. As long as we use the `DEV` or `STAGING` environment this should be fine imo. These environments are meant to be spammed with non-sense and clean ups should be in there own responsibility.

There is a point to be made about mocking the backend though. Independence. Our integration tests would still work when the backend is down or broken. That can be helpful, particularly with a flaky kind of backend like the one we're dealing with atm.

The drawback of course, is that it's not *really* an integration test anymore because it doesn't test the backend and 3L's integration with it anymore.

danielgranhao: I see your point, but still think we should make it easier for the clean-ups to not remove channel states from test wallets we want to be a bit longer lived. These very short-lived wallets created in int tests could have appended to the 3L version some code that distinguishes them. WDYT?

> The drawback of course, is that it's not really an integration test anymore because it doesn't test the backend and 3L's integration with it anymore.

It would still be an integration test testing the integration of `eel` and `mole`. Just not the integration with the backend :)

danielgranhao: Let's deal with this in another PR

danielgranhao (approved):

PR #213, opened 2023-02-09
Set timeouts for gRPC client

andrei-21:
Such approach also will allow us to add TLS server authentication.

### DIFF
--- a/eel/src/lsp_client/mod.rs
+++ b/eel/src/lsp_client/mod.rs
@@ -46,8 +46,8 @@ impl LspClient {
         })
     }
 
-    // The function must be async because `Endpoint::connect_lazy()` access
-    // aync runtime.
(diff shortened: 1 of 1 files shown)


eel/src/lsp_client/mod.rs
```diff
@@ -23,41 +27,46 @@ impl LspClient {
...
     }
 
-    fn wrap_request<T>(&self, request: T) -> Request<T> {
-        let mut request = Request::new(request);
-        request
-            .metadata_mut()
-            .insert("authorization", self.bearer.clone());
-        request
+    // The function must be async because `Endpoint::connect_lazy()` access
+    // aync runtime.
```
danielgranhao, line 50:
```suggestion
    // The function must be async because `Endpoint::connect_lazy()` accesses
    // the async runtime.
```

danielgranhao (approved):

PR #223, opened 2023-02-14
Add integrationtests for 3L

gcomte:


tests/node_info_test.rs
```diff
@@ -0,0 +1,124 @@
...
+            let node = LightningNode::new(self.config.clone(), Box::new(events_handler));
+
+            // Wait for the the P2P background task to connect to the LSP
+            sleep(Duration::from_millis(1500));
+
+            node
+        }
+    }
+
+    pub struct PrintEventsHandler {}
```
danielgranhao, line 85: I think you can also get this from `eel` using the `#[path = ""]` approach.

gcomte: Yes! :muscle: DOWN WITH DUPLICATE CODE

gcomte: Looks like this cannot be easily done, because the `PrintEventsHandler` in `eel` implements `EventHandler` while this `PrintEventsHandler` here in 3L is required to implement `EventsCallback`, which is being wrapped into an `EventHandler` within the 3L library.

Interestingly enough both `EventHandler` and `EventsCallback` both define the same methods. Not sure why the encapsulation but I think it has to do with error mapping.

danielgranhao: Yes, that's the reason for the encapsulation. Well.. seems like we can't remove the duplicate code in this case 😢

Makefile
```diff
@@ -26,7 +26,8 @@ test:
 integrationtests: FILE = *
 integrationtests: TEST = ''
 integrationtests:
-	cargo test --manifest-path eel/Cargo.toml --features nigiri --test '$(FILE)' -- $(TEST)
+	LSPD_HOME=lspd RGS_HOME=rgs cargo test --features nigiri --test '$(FILE)' -- $(TEST)
```
danielgranhao, line 29: The env variables also need to be set up in the Makefile inside eel :)

Makefile
```diff
@@ -26,7 +26,8 @@ test:
...
+	#LSPD_HOME=../lspd RGS_HOME=../rgs cargo test --manifest-path eel/Cargo.toml --features nigiri --test '$(FILE)' -- $(TEST)
```
danielgranhao, line 30: Do we need this comment?
```suggestion
```

gcomte: Sorry, I should not have commited this.

That's not a comment but the actual instruction to run the eel integration tests. I just commented it for local testing and then accidentally commited it.

Fixed and force-pushed rn.

danielgranhao (approved):

PR #227, opened 2023-02-15
Stop payment of recognized invoices

danielgranhao:


gcomte (commented):
What happens if an invoice was already scanned and almost paid, but the transaction failed? Is the invoice then also 'Already Used'?

danielgranhao:
> What happens if an invoice was already scanned and almost paid, but the transaction failed? Is the invoice then also 'Already Used'?

Yes. In the future, we can improve to allow further attempts to pay an invoice for which a previous attempt has failed, but for now, this is a good enough solution to start.

gcomte (approved):

PR #231, opened 2023-02-20
Add cargo workspace

andrei-21:


danielgranhao (approved):
Cool! Does running `cargo test --workspace ...` also run the tests on 3L?

PR #237, opened 2023-02-22
Colored output for the example node

andrei-21:
image

### DIFF
--- a/eel/examples/node/cli.rs
+++ b/eel/examples/node/cli.rs
@@ -23,7 +23,7 @@ pub(crate) fn poll_for_user_input(node: &LightningNode, log_file_path: &str) {
     let history_path = Path::new("./.ldk/cli_history");
     let _ = rl.load_history(history_path);
 
-    let prompt = "eel ⚡ ".bold().yellow().to_string();
+    let prompt = "eel ϟ ".bold().yellow().to_string();
     loop {
         let line = match rl.readline(&prompt) {
             Ok(line) => line,


eel/examples/node/cli.rs
```diff
@@ -22,12 +23,13 @@ pub(crate) fn poll_for_user_input(node: &LightningNode, log_file_path: &str) {
     let history_path = Path::new("./.ldk/cli_history");
     let _ = rl.load_history(history_path);
 
+    let prompt = "eel ⚡ ".bold().yellow().to_string();
```
danielgranhao, line 26: 😎

danielgranhao (commented):
Cool! I noticed something weird though.. When using the backspace, the whole command moves to the left by 1 char 🤔

danielgranhao:
> > Cool! I noticed something weird though.. When using the backspace, the whole command moves to the left by 1 char thinking
> 
> Can you elaborate please, I cannot observe anything like this?

Sure. I've just checked and this happens on my IDE's terminal but not on the OS default terminal. Maybe that's why you don't experience it.

If I type `test` I see:
```
eel ⚡  test
```
Notice that there's 2 spaces between `⚡` and `t`. After a backspace:
```
eel ⚡ tes 
```
And if I continue to write, e.g. writing test after the `tes`:
```
eel ⚡ tes test
```
A space between `tes` and `test` is shown but if I execute the command it's not actually there. It's just some weird graphical bug, but an annoying one.

danielgranhao (approved):

PR #242, opened 2023-02-23
Update exchange rates

andrei-21:
```sh
GRAPHQL_API_URL=*** make run-3l ARGS=CHF
```
will change currency for the example node.

### DIFF
--- a/eel/src/interfaces.rs
+++ b/eel/src/interfaces.rs
@@ -31,7 +31,7 @@ pub trait EventHandler: Send + Sync {
 }
 
 #[derive(Clone)]
-pub struct ExchangeRate {
+pub struct ExchangeRates {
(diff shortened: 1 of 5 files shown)


eel/src/interfaces.rs
```diff
@@ -30,6 +30,12 @@ pub trait EventHandler: Send + Sync {
     fn payment_failed(&self, payment_hash: String) -> Result<()>;
 }
 
+#[derive(Clone)]
+pub struct ExchangeRate {
```
danielgranhao, line 34:
```suggestion
pub struct ExchangeRates {
```

danielgranhao (approved):
Just a small nit. Feel free to keep it as is if you disagree.

PR #245, opened 2023-02-24
    Pass invoice payer as event handler

andrei-21:


gcomte (commented):
When already implementing this, we could also fail it if `payment_failed_permanently` is `true`, because it seems to be more obvious of a failing case than `all_paths_failed`, right?

Or is the abandoning done automatically in the background in that case?

eel/src/event_handler.rs
```diff
@@ -203,7 +203,20 @@ impl EventHandler for LipaEventHandler {
...
+                network_update: _,
+                all_paths_failed,
+                ..
+            } => {
+                if all_paths_failed {
+                    if let Some(payment_id) = payment_id {
+                        self.channel_manager.abandon_payment(payment_id);
+                    }
+                }
+            }
```
danielgranhao, line 219: This may not be what we want to do. The `InvoicePayer` (see [docs](https://docs.rs/lightning-invoice/0.21.0/lightning_invoice/payment/index.html)) should be capable of reacting to failed payment paths and retrying a few times before failing. 
What is being done here may conflict with its logic. Maybe we have a problem with the `InvoicePayer` setup?

danielgranhao (approved):

PR #246, opened 2023-02-27
Fix comment in UDL file

gcomte:


danielgranhao:
I think we can even remove this error code. We don't expose the sync functionality to the consumer, right?

PR #247, opened 2023-02-27
Remove get_object functionality from storage trait

gcomte:


eel/src/storage_persister.rs
```diff
@@ -263,34 +263,28 @@ impl StoragePersister {
...
-                    let mut buffer = Cursor::new(&data);
-                    let (block_hash, channel_manager) = <(
+                        "Failed to read a remote ChannelManager",
+                    )?;
+
+                if encrypted_data.is_empty() {
+                    Err(permanent_failure(
+                        "Failed to find remote ChannelManager even though this was determined to be a Recovery start",
+                    ))?;
+                }
```
danielgranhao, line 278: If there's no `ChannelManager` we won't get an empty vec, but an runtime error with `RuntimeErrorCode::ObjectNotFound`, right?

gcomte: Right, thanks! The idea was to do something like this: ba437dc866bfb70f862ebf34cd351ae5f9bc4c0f

Is there a way to make this more concise?

gcomte: Exactly what I was looking for, thank you :pray:

danielgranhao (approved):

PR #250, opened 2023-03-02
Switch to asynchronous remote monitor persistence

danielgranhao:


eel/src/storage_persister.rs
```diff
@@ -416,33 +413,32 @@ impl<ChannelSigner: WriteableEcdsaChannelSigner> Persist<ChannelSigner> for Stor
...
-        match std::thread::spawn(move || {
-            sync_persist_monitor_remotely(storage, &encryption_key, data, channel_id, retries)
-        })
-        .join()
-        {
-            Ok(Ok(_)) => ChannelMonitorUpdateStatus::Completed,
-            Ok(Err(_)) => ChannelMonitorUpdateStatus::PermanentFailure,
-            Err(_) => ChannelMonitorUpdateStatus::PermanentFailure,
-        }
+        // Launch thread instead of tokio task due to the use of blocking reqwest in our
```
gcomte, line 428: the use of blocking reqwest in our [???]

gcomte (commented):
Nice, code base looks a lot simpler now.

I'd call the PR `Switch to multithreaded remote monitor persistence` because in rust the async word usually refers to [asynchronous, single-threaded], which is not the case here.

eel/src/storage_persister.rs
danielgranhao: It continues on the next line :)

danielgranhao:
> I'd call the PR Switch to multithreaded remote monitor persistence because in rust the async word usually refers to [asynchronous, single-threaded], which is not the case here.

I think the current name makes sense. The adjective "async" applies to the general remote monitor persistence approach. From the point of view of LDK, monitors get persisted asynchronously. LDK asks us to persist a monitor, we return "immediately" and only later we let LDK know the monitor got persisted.

danielgranhao:
I switched "async" with "asynchronous" to help avoid mixing the concept of something being async with the rust `async` keyword.

eel/src/storage_persister.rs
gcomte: Oh, I'm sorry, this was poorly formatted in my review screen

PR #255, opened 2023-03-13
Add periodic GitHub actions check

gcomte:
Check every day at 2:21 AM UTC for environmental changes such as new Rust version released, found security vulnerabilities, interfaces of remote services changed (-> integration tests), etc.

This becomes particularly relevant if we don't actively work on the project for a long time.


danielgranhao (approved):

gcomte:
> We can skip weekends because we most likely would not react on that.

Yeah it's also likely that there is less change in the environment, because other people don't work either. I applied your recommendation.

PR #307, opened 2023-03-30
Implement payment db upgrade strategy

danielgranhao:
Builds on top of #305 to avoid already creating a new schema version.


eel/src/payment_store.rs
```diff
@@ -4,13 +4,15 @@ use crate::interfaces::ExchangeRates;
...
-use perro::{MapToError, OptionToError};
+use perro::{permanent_failure, MapToError, OptionToError};
 use rusqlite::types::Type;
-use rusqlite::{Connection, Row};
+use rusqlite::{Connection, OptionalExtension, Row};
 use std::convert::TryFrom;
 use std::str::FromStr;
 use std::time::SystemTime;
 
+const SCHEMA_VERSION: u64 = 1;
```
gcomte, line 14: `u64` ? This allows for a looooot of schema versions :joy:

danielgranhao: I didn't see a reason to restrict it, but we can go with a smaller integer type. Changed to `u16`.

PR #336, opened 2023-04-25
Decrease on-chain foregroung sync to 5 minutes

andrei-21:


gcomte (approved):

gcomte:
Hmmm. I'm not a big fan of writing production code that is only being required for testing. How about we add these fields as Optionals to the use `eel::config::Config`, and if they are not set in the config, we default to the hard-coded values ?

gcomte:
Well. We don't necessarily have to expose it to the library consumer, but only to `3L`.

So the settings could be hard-coded in `3L` as weel as in the `eel`-test.

PR #379, opened 2023-05-19
Enable backtrace and debug symbols for our and LDK crates

andrei-21:
It increases the size of 3l-node example node from 28M to 81M.


gcomte (approved):

danielgranhao:
Should we add `enable_backtrace()` to the other top-level functions?

PR #385, opened 2023-05-23
Interface for sats and fiat amounts

andrei-21:
I suggest to replace all msats and fiat values with `Amount` in the UDL file.
In rust to round and convert from msats  we can use the following construction:
```rust
       let channel_minimum_fee = fee
            .channel_minimum_fee_msat
            .to_amount_up(self.get_exchange_rate());
```

### DIFF
--- a/src/amount.rs
+++ b/src/amount.rs
@@ -4,7 +4,7 @@ use std::time::SystemTime;
 pub struct FiatValue {
     pub minor_units: u64,
     pub currency_code: String,
-    pub updated_at: SystemTime,
+    pub converted_at: SystemTime,
 }
 
 pub struct Amount {
@@ -45,7 +45,7 @@ fn msats_to_amount(rounding: Rounding, msats: u64, rate: Option<ExchangeRate>) -
     let fiat = rate.map(|rate| FiatValue {
         minor_units: round(msats * 100 / rate.rate as u64, rounding),
         currency_code: rate.currency_code,
-        updated_at: rate.updated_at,
+        converted_at: rate.updated_at,
     });
     Amount { sats, fiat }
 }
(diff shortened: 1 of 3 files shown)


src/amount.rs
```diff
@@ -0,0 +1,118 @@
+use eel::interfaces::ExchangeRate;
+use std::time::SystemTime;
+
+pub struct FiatValue {
+    pub minor_units: u64,
+    pub currency_code: String,
+    pub updated_at: SystemTime,
```
danielgranhao, line 7: The name `updated_at` might lead to the assumption that an `Amount` could have its fiat value updated as time goes by. I would prefer:
```suggestion
    pub converted_at: SystemTime,
```

src/lipalightninglib.udl
```diff
@@ -314,13 +314,24 @@ dictionary FiatValues {
...
 dictionary ExchangeRate {
     string currency_code;
     u32 rate;
     timestamp updated_at;
 };
 
+dictionary Amount {
+    u64 sats;
+	FiatValue? fiat;
+};
```
danielgranhao, line 333:
```suggestion
dictionary Amount {
    u64 sats;
    FiatValue? fiat;
};
```

src/lipalightninglib.udl
```diff
@@ -314,13 +314,24 @@ dictionary FiatValues {
     u64 amount_usd;
 };
 
+dictionary FiatValue {
+    u64 minor_units;
+    string currency_code;
+	timestamp updated_at;
+};
```
danielgranhao, line 321:
```suggestion
dictionary FiatValue {
    u64 minor_units;
    string currency_code;
    timestamp updated_at;
};
```

danielgranhao (approved):

PR #388, opened 2023-05-24
Expose get_payment_amount_limits() in UDL

gcomte:
This was missing in https://github.com/getlipa/lipa-lightning-lib/pull/380

### DIFF
--- a/src/lipalightninglib.udl
+++ b/src/lipalightninglib.udl
@@ -40,6 +40,12 @@ interface LightningNode {
     [Throws=LnError]
     LspFee query_lsp_fee();
 
+    // Get the current limits for the amount (in SAT) that can be transferred in a single payment. Currently there are only limits for receiving payments.
+    // The limits (partly) depend on the channel situation of the node, so it should be called again every time the user is about to receive a payment.
+    // The limits stay the same regardless of what amount wants to receive (= no changes while he's typing the amount)
+    [Throws=LnError]
+    PaymentAmountLimits get_payment_amount_limits();
+
     // Calculate the actual LSP fee in millisats for the given amount of an incoming payment.
     // If the already exisitng inbound capacity is enough, no new channel is required: The method returns 0.
     //    - amount_msat - amount in millisats to compute LSP fee for


src/lipalightninglib.udl
```diff
@@ -40,6 +40,11 @@ interface LightningNode {
     [Throws=LnError]
     LspFee query_lsp_fee();
 
+    // Limits in the amount sats a payment is allowed to transfer. Currently there are only limits for receiving payments.
```
danielgranhao, line 43:
```suggestion
    // Get the current limits for the number of sats that can be transferred in a single payment. Currently, there are only limits for receiving payments.
```

src/lipalightninglib.udl
```diff
@@ -40,6 +40,11 @@ interface LightningNode {
...
+    // The limits (partly) depend on the channel situation of the node, so it should be called again every time the user is about to receive a payment.
```
danielgranhao, line 44:
```suggestion
    // The limits (partly) depend on the channel situation of the node, so it should be called every time the user is about to issue an invoice.
```

src/lipalightninglib.udl
```diff
@@ -40,6 +40,11 @@ interface LightningNode {
...
+    // The limits stay the same regardless of what amount the user enters that he wants to receive; they solely depend on the channel situation.
```
danielgranhao, line 45:
```suggestion
    // The limits stay the same regardless of what amount the user wants to receive; they solely depend on the channel situation.
```

src/lipalightninglib.udl
gcomte: That change doesn't cover spontaneous payments anymore .. (I'm particularly thinking of the use case when a payment has to be pulled down via LNURLw)

src/lipalightninglib.udl
gcomte: I agree my phrase reads badly. But I purposefully wanted to write down "entering amount" because it was a FAQ, whether the limits change while the user is typing an amount.

src/lipalightninglib.udl
danielgranhao: An invoice still have to be issued when using LNURLw. But I agree that the user of this library might not realize it. I proposed the change because the original version makes it seem like this should be used right before the payment was received and not before the invoice was issued.

src/lipalightninglib.udl
danielgranhao: To me, it remains understandable, but we can keep it if you disagree.

gcomte: How about https://github.com/getlipa/lipa-lightning-lib/compare/b6fdc36f2795b358153e9087bece611928a69935..8bb0a178bb02b98e1518fd0aa7382316a2e287ea ?

src/lipalightninglib.udl
gcomte: From a UX point of view there is no creation of any invoice.

src/lipalightninglib.udl
gcomte: Yes, much better! Check out https://github.com/getlipa/lipa-lightning-lib/compare/8bb0a178bb02b98e1518fd0aa7382316a2e287ea..81b2ff653a7f0ba2de31b8f900c2eaa1d432b4cd

danielgranhao (approved):

PR #394, opened 2023-05-26
Amount interface for `InvoiceDetails`

andrei-21:


examples/3l-node/cli.rs
```diff
@@ -138,7 +138,7 @@ fn setup_editor(history_path: &Path) -> Editor<CommandHinter, DefaultHistory> {
     hints.insert(CommandHint::new("nodeinfo", "nodeinfo"));
     hints.insert(CommandHint::new("lspfee", "lspfee"));
     hints.insert(CommandHint::new(
-        "calculatelspfee <amount in millisat>",
+        "calculatelspfee <amount in sat>",
```
danielgranhao, line 141:
```suggestion
        "calculatelspfee <amount in sats>",
```

examples/3l-node/cli.rs
```diff
@@ -162,7 +162,7 @@ fn setup_editor(history_path: &Path) -> Editor<CommandHinter, DefaultHistory> {
     ));
     hints.insert(CommandHint::new("payinvoice <invoice>", "payinvoice "));
     hints.insert(CommandHint::new(
-        "payopeninvoice <invoice> <amount in msat>",
+        "payopeninvoice <invoice> <amount in sat>",
```
danielgranhao, line 165:
```suggestion
        "payopeninvoice <invoice> <amount in sats>",
```

examples/3l-node/cli.rs
```diff
@@ -182,17 +182,17 @@ fn setup_editor(history_path: &Path) -> Editor<CommandHinter, DefaultHistory> {
 fn help() {
     println!("  nodeinfo");
     println!("  lspfee");
-    println!("  calculatelspfee <amount in millisat>");
+    println!("  calculatelspfee <amount in sat>");
```
danielgranhao, line 185:
```suggestion
    println!("  calculatelspfee <amount in sats>");
```

src/invoice_details.rs
```diff
@@ -0,0 +1,185 @@
...
+    pub description: String,
+    pub payment_hash: String,
+    pub payee_pub_key: String,
+    pub creation_timestamp: SystemTime,
+    pub expiry_interval: Duration,
+    pub expiry_timestamp: SystemTime,
+}
+
+impl InvoiceDetails {
+    pub(crate) fn from_our_invoice(invoice: Invoice, rate: &Option<ExchangeRate>) -> Self {
```
danielgranhao, line 19: If you prefer "our", feel free to reject the suggestion.
```suggestion
    pub(crate) fn from_locally_issued_invoice(invoice: Invoice, rate: &Option<ExchangeRate>) -> Self {
```

src/lipalightninglib.udl
```diff
@@ -253,7 +253,7 @@ interface LiquidityLimit {
 // Information embedded in an invoice
 dictionary InvoiceDetails {
     string invoice; // The BOLT11 invoice.
-    u64? amount_msat; // Payment amount in msats, if specified. If not available, invoice is an open-amount invoice
+    Amount? amount; // Payment amount, if specified. If not available, invoice is an open-amount invoice
```
danielgranhao, line 256: Now that `InvoiceDetails` carries a fiat value, we need to make clear which exchange rate is used for that fiat value. At the moment, I'm worried about the following situation in which a discrepancy can arise:
1. Consumer creates an invoice and looks at the fiat value contained within `InvoiceDetails`
    * It is reasonable for the consumer to expect that this is the fiat value that will get persisted in the local records
2. Later, the consumer gets the list of `Payments` and the `InvoiceDetails` inside the related payment has a different fiat value.

Maybe we don't need to worry about this edge case

danielgranhao (approved):
Generally LGTM. Left some very minor suggestions. 

Maybe one

examples/3l-node/cli.rs
gcomte: We've been very inconsistent with that. Myself I would prefer `SAT` in capital letters (for plural and singular). This is also how it's being displayed in the Lipa Wallet.

src/lipalightninglib.udl
```diff
@@ -253,8 +253,12 @@ interface LiquidityLimit {
 // Information embedded in an invoice
 dictionary InvoiceDetails {
     string invoice; // The BOLT11 invoice.
-    u64? amount_msat; // Payment amount in msats, if specified. If not available, invoice is an open-amount invoice
-        // and the user should be promted for how much they want to pay.
+	// Payment amount, if specified. If not available, invoice is an open-amount invoice
+    // and the user should be promted for how much they want to pay.
+	// The fiat value is calculated base on the "natural" exchange rate:
+	//  - for a new invoice current exchange rate is used
+	//  - for old invoices historic values are used
```
danielgranhao, line 260:
```suggestion
    // Payment amount, if specified. If not available, invoice is an open-amount invoice
    // and the user should be promted for how much they want to pay.
    // The fiat value is calculated base on the "natural" exchange rate:
    //  - for a new invoice current exchange rate is used
    //  - for old invoices historic values are used
```

danielgranhao (approved):

PR #395, opened 2023-05-26
Enable MPP

gcomte:
- Sets enabling multipath-payments flag in invoices
- Tests receiving mpp
- Tests sending mpp

### DIFF
--- a/eel/tests/receiving_payments_test.rs
+++ b/eel/tests/receiving_payments_test.rs
@@ -182,7 +182,7 @@ mod receiving_payments_test {
             assert!(payment_amount > ONE_M_SATS); // Biggest channel is 1M sats; This payment involves transferring through more than 1 channel
             let invoice = issue_invoice(
                 &node,
-                node.get_node_info().channels_info.inbound_capacity_msat - TWO_HUNDRED_K_SATS,
+                payment_amount,
             );
 
             wait_for!(nigiri::lnd_pay_invoice(NodeInstance::LspdLnd, &invoice).is_ok());
--- a/eel/tests/receiving_payments_test.rs
+++ b/eel/tests/receiving_payments_test.rs
@@ -167,7 +167,7 @@ mod receiving_payments_test {
 
             // Add 2M sats of inbound capacity - More than the max allowed receive amount
             let node_id = node.get_node_info().node_pubkey.to_hex();
-            assert_eq!(nigiri::get_number_of_txs_in_mempool(), Ok::<u64, String>(2)); // 2 times jit channel open flow
+            assert_eq!(nigiri::get_number_of_txs_in_mempool(), Ok::<u64, String>(2)); // 2 times jit channel open flow (zero-conf)
             let _ = nigiri::lnd_node_open_channel(NodeInstance::LspdLnd, &node_id, false).unwrap(); // 1M sats
             let _ = nigiri::lnd_node_open_channel(NodeInstance::LspdLnd, &node_id, false).unwrap(); // 1M sats
             wait_for_eq!(nigiri::get_number_of_txs_in_mempool(), Ok::<u64, String>(4));
--- a/eel/tests/receiving_payments_test.rs
+++ b/eel/tests/receiving_payments_test.rs
@@ -22,7 +22,10 @@ mod receiving_payments_test {
     const TWO_K_SATS: u64 = 2_000_000;
     const TEN_K_SATS: u64 = 10_000_000;
     const TWENTY_K_SATS: u64 = 20_000_000;
+    const TWO_HUNDRED_K_SATS: u64 = 200_000_000;
     const HALF_M_SATS: u64 = 500_000_000;
+    const ONE_M_SATS: u64 = 1_000_000_000;
+    const TWO_M_SATS: u64 = 2_000_000_000;
 
     // The amount of sats the LSP provides to the user as inbound capacity.
     // See https://github.com/getlipa/lipa-lightning-lib/blob/b821162df982799c497e083e9707aa421aee43a8/lspd/compose.yaml#LL63C44-L63C46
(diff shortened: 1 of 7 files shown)


eel/tests/receiving_payments_test.rs
```diff
@@ -164,14 +167,27 @@ mod receiving_payments_test {
...
+            let payment_amount = inbound_capacity - TWO_HUNDRED_K_SATS; // todo use full inbound capacity as soon as the number is accurate
+            assert!(payment_amount > ONE_M_SATS); // Biggest channel is 1M sats; This payment involves transferring through more than 1 channel
+            let invoice = issue_invoice(
+                &node,
+                node.get_node_info().channels_info.inbound_capacity_msat - TWO_HUNDRED_K_SATS,
+            );
+
+            wait_for!(nigiri::lnd_pay_invoice(NodeInstance::LspdLnd, &invoice).is_ok());
+
+            assert!(node.get_node_info().channels_info.local_balance_msat > TWO_M_SATS);
```
danielgranhao, line 190: Why not check the exact amount?
```suggestion
            assert_eq!(node.get_node_info().channels_info.local_balance_msat, initial_balance + payment_amount);
```
Suggestion needs `initial_balance` to be set of course.

eel/tests/receiving_payments_test.rs
```diff
@@ -164,14 +167,27 @@ mod receiving_payments_test {
...
 
-            // todo also test multipath payments
+            // test receiving MPP
+            let inbound_capacity = node.get_node_info().channels_info.inbound_capacity_msat;
+            let payment_amount = inbound_capacity - TWO_HUNDRED_K_SATS; // todo use full inbound capacity as soon as the number is accurate
+            assert!(payment_amount > ONE_M_SATS); // Biggest channel is 1M sats; This payment involves transferring through more than 1 channel
+            let invoice = issue_invoice(
+                &node,
+                node.get_node_info().channels_info.inbound_capacity_msat - TWO_HUNDRED_K_SATS,
+            );
```
danielgranhao, line 186:
```suggestion
            let invoice = issue_invoice(
                &node,
                payment_amount,
            );
```

eel/tests/receiving_payments_test.rs
```diff
@@ -164,14 +167,27 @@ mod receiving_payments_test {
 
             // Add 2M sats of inbound capacity - More than the max allowed receive amount
             let node_id = node.get_node_info().node_pubkey.to_hex();
+            assert_eq!(nigiri::get_number_of_txs_in_mempool(), Ok::<u64, String>(2)); // 2 times jit channel open flow
```
danielgranhao, line 170: Probably this is obvious but I was getting confused why the already open channel would still have TXs in the mempool :) 
```suggestion
            assert_eq!(nigiri::get_number_of_txs_in_mempool(), Ok::<u64, String>(2)); // 2 times jit channel open flow (zero-conf)
```

eel/tests/sending_payments_test.rs
```diff
@@ -105,12 +113,51 @@ mod sending_payments_test {
...
+            .unwrap();
+        assert!(invoice_details.invoice.starts_with("lnbc"));
+
+        wait_for!(nigiri::pay_invoice(LspdLnd, &invoice_details.invoice).is_ok());
+        assert_eq!(
+            node.get_node_info().channels_info.local_balance_msat,
+            NINE_HUNDRED_K_SATS + NINE_HUNDRED_K_SATS + FOURTY_K_SATS
+        );
+
+        // Node has 2 channels of 2M SAT each. Paying 1.5M SAT requires sending through both of them
```
danielgranhao, line 154:
```suggestion
        // Node has 2 channels of 1M SAT each. Paying 1.5M SAT requires sending through both of them
```

eel/tests/sending_payments_test.rs
```diff
@@ -105,12 +113,51 @@ mod sending_payments_test {
...
+        let invoice = nigiri::issue_invoice(LspdLnd, "MPP", MORE_THAN_ONE_M_SATS, 3600).unwrap();
+
+        node.pay_invoice(invoice, String::new()).unwrap();
+
+        wait_for!(nigiri::query_node_balance(LspdLnd).unwrap() > MORE_THAN_ONE_M_SATS);
```
danielgranhao, line 159: A suggested on the receiving test, we can check for exact values. Also, does it need to be a wait instead of assert?
```suggestion
        wait_for_eq!(nigiri::query_node_balance(LspdLnd).unwrap(), initial_balance + MORE_THAN_ONE_M_SATS);
```

eel/tests/receiving_payments_test.rs
gcomte: Yes, thanks

eel/tests/receiving_payments_test.rs
gcomte: Yes it was obvious to me. Thanks for pointing out. Elaborating makes sense indeed

eel/tests/sending_payments_test.rs
gcomte: Check out e6fe545c4a805c42b5b94372f385beac37611cd6

eel/tests/receiving_payments_test.rs
gcomte: Check out e6fe545c4a805c42b5b94372f385beac37611cd6

eel/tests/sending_payments_test.rs
gcomte: Check out e1b5bb95f72c44629f9ec2a1aa87b128692e3a7c

Yes, the wait is required because the payment happens in another thread, so the assert would be called before the payment actually went through.

danielgranhao: Ah yes, you're right! Thanks for elaborating

danielgranhao (approved):

danielgranhao (approved):

PR #402, opened 2023-05-30
Query lsp info sooner after failure

danielgranhao:


eel/src/task_manager.rs
```diff
@@ -179,8 +179,12 @@ impl TaskManager {
                                 error!("Connecting to peer {} failed: {}", peer, e);
                             }
                         }
+                        Some(period)
+                    }
+                    Err(e) => {
+                        error!("Failed to query LSP, retrying in 10 seconds: {}", e);
+                        Some(Duration::from_secs(10))
```
gcomte, line 186: How did you come up with 10 seconds? Feels pretty long as well.
Also let's put this at least into a constant, if we don't make it configurable (yet).

eel/tests/p2p_connection_test.rs
```diff
@@ -31,23 +33,26 @@ mod p2p_connection_test {
...
-            wait_for_eq!(node.get_node_info().num_peers, 0);
-        }
+        assert_eq!(
+            node.query_lsp_fee(),
+            Err(runtime_error(
+                RuntimeErrorCode::LspServiceUnavailable,
+                "Failed to get LSP info"
+            ))
+        );
+        wait_for_eq!(node.get_node_info().num_peers, 0);
```
gcomte, line 48: Why do you wait for that? There shouldn't have been any peer since the node was started, right?

gcomte, line 48: Would it be possible to move this code block in front of `nigiri::ensure_environment_running();` such that `lspd` does not need pausing? Or is there something else required from the environment such that the node couldn't be started?

eel/src/task_manager.rs
danielgranhao: It's what we use for the `reconnect_to_lsp` task. I thought it made sense to be the same, but I agree 10 secs be long. Changed to 5 and created `PeriodConfig` so that it becomes defined together with other periods.

eel/tests/p2p_connection_test.rs
danielgranhao: Yes, it something that remained from the previous code. Fixed!

danielgranhao: Not sure I understand what you propose. The goal is to start the node without having lspd running. If we start the node before `nigiri::ensure_environment_running();`, we don't know if lspd is running or not.

gcomte (approved):

eel/tests/p2p_connection_test.rs
gcomte: Fixed: https://github.com/getlipa/lipa-lightning-lib/compare/40c9a8aed3598b56cd2277cb3c6f7e04e819e93d..92a9f39937a2a6e383be57c891e6a3ab6b274fc3

PR #409, opened 2023-06-02
Store exchange rates instead of fiat values

andrei-21:


danielgranhao:
Looks good! Just please fix the several places where tabs are used instead of spaces.

danielgranhao (approved):

PR #411, opened 2023-06-02
Add network to `InvoiceDetails`

danielgranhao:
I realized there's no way for the consumer of 3L to fulfil the obligation of providing same network invoices to `LightningNode::pay_invoice()`. With these changes, the consumer can learn the network of an invoice when calling `LightningNode::decode_invoice()`.


danielgranhao:
Then the consumer can't present the user with the network the invoice actually belongs to. That could be useful.

PR #420, opened 2023-06-08
Restructure Error model

gcomte:
This is just a draft to see how we can change the model and how much work it entails.


src/lipalightninglib.udl
```diff
@@ -76,25 +76,25 @@ interface LightningNode {
...
     // Additional Parameters:
     //    - amount_sat - amount in sats to be paid
-    [Throws=LnError]
+    [Throws=PayError]
     void pay_open_invoice(string invoice, u64 amount_sat, string metadata);
 
     // Get a list of the latest payments
     //    - number_of_payments - the maximum number of payments that will be returned
-    [Throws=LnError]
+    [Throws=PayError]
```
danielgranhao, line 92: Maybe this is just due to the WIP nature of the PR, but why would `PayError` be the type here? The same question would apply to `get_payment()` and `create_invoice()`.

gcomte: Yes that was a bit prototype and I didn't know how I wanted to split the errros at first. Should be fixed now.

src/lipalightninglib.udl
```diff
@@ -405,20 +405,28 @@ enum RuntimeErrorCode {
...
+    InvalidInput(string msg);
+    RuntimeError(PayErrorCode code, string msg);
+    PermanentFailure(string msg);
+};
+
+// A code that specifies the PayError that occurred
+enum PayErrorCode {
     "InvoiceNetworkMismatch", // The invoice belongs to different network. Please use a different one.
     "AlreadyUsedInvoice", // An already recognized invoice tried to be paid. Either a payment attempt is in progress or the invoice has already been paid.
     "PayingToSelf", // A locally issued invoice tried to be paid. Self-payments are not supported.
```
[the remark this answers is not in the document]
gcomte: Check out 050b3f7e6e0188498d89190053a1143ba14057dc for your first point.
With this we lose the error types, which I think would be quite useful for the library consumer to appropriately inform the user about what exactly is wrong with his invoice. Do we need specific types for `invalid_input` as well? Also removing these errors decreases the amount of possible error cases for `PayError` significantly, which then brings up the question of whether it even needs to be split into its own error type.

I don't really understand what you mean with the second point and why returning an `invalid_input` wouldn't make sense here.

PR #422, opened 2023-06-08
Fix error mapping for `PaymentExpired` and `DuplicatePayment`

andrei-21:


eel/src/lib.rs
```diff
@@ -494,34 +493,26 @@ impl LightningNode {
...
+                    "Invoice has expired",
+                )),
+                RetryableSendFailure::RouteNotFound => Err(runtime_error(
+                    RuntimeErrorCode::NoRouteFound,
+                    "Failed to find any route",
+                )),
+                RetryableSendFailure::DuplicatePayment => Err(runtime_error(
+                    RuntimeErrorCode::AlreadyUsedInvoice,
+                    "Already used invoice",
+                )),
```
danielgranhao, line 514: Our current payment tracking mechanism should prevent this case from happening. Maybe return a permanent error instead?

danielgranhao (approved):

PR #436, opened 2023-06-22
Remove unused error codes from `RuntimeErrorCode`

andrei-21:
Remove `ObjectNotFound`, `NonExistingWallet`, `GenericError`

### DIFF
--- a/eel/src/storage_persister.rs
+++ b/eel/src/storage_persister.rs
@@ -30,7 +30,7 @@ use lightning::util::ser::{ReadableArgs, Writeable};
 use lightning_persister::FilesystemPersister;
 use log::{debug, error, warn};
 use perro::Error::RuntimeError;
-use perro::{invalid_input, permanent_failure, runtime_error, MapToError};
+use perro::{invalid_input, permanent_failure, MapToError, ResultTrait};
 use std::fs;
 use std::io::{BufReader, Cursor};
 use std::ops::Deref;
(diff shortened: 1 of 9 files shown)


eel/src/task_manager.rs
```diff
@@ -181,13 +181,13 @@ impl TaskManager {
...
                             if let Err(e) = connect_peer(&peer, peer_manager).await {
-                                error!("Connecting to peer {} failed: {}", peer, e);
+                                error!("Connecting to peer {peer} failed: {e}");
                             }
                         }
                         Some(config.success_period)
                     }
                     Err(e) => {
-                        error!("Failed to query LSP, retrying in 10 seconds: {}", e);
+                        error!("Failed to query LSP, retrying in 10 seconds: {e}");
```
danielgranhao, line 190: Just to take advantage of this line being edited:
```suggestion
                        error!(
                            "Failed to query LSP, retrying in {} seconds: {e}",
                            config.failure_period.as_secs()
                        );
```

eel/src/storage_persister.rs
```diff
@@ -318,22 +318,9 @@ impl StoragePersister {
...
-                ..
-            }) => {
-                return Err(runtime_error(
-                    RuntimeErrorCode::NonExistingWallet,
-                    "Failed to find remote ChannelManager",
-                ))
-            }
-            Err(e) => return Err(e),
-        };
+            .get_object(OBJECTS_BUCKET.to_string(), MANAGER_KEY.to_string())?;
```
danielgranhao, line 323: If the recovery method fails, how can the consumer distinguish between a non-existing wallet and the remote storage service being unavailable?

danielgranhao: The idea was that the `RemoteStorage::get_object()` implementation would return a specific error when the remote storage could be reached, but it reported no such object. Looking at codebase at `main` I think we had a bug as we only used `ObjectNotFound` in the file storage implementation, and not on the proper remote storage implementation (`eel_interface_impl.rs`).

danielgranhao (approved):
Nice!

danielgranhao (approved):

PR #503, opened 2023-07-20
Ignore our deps in dependabot

danielgranhao:


danielgranhao:
We do, but almost every time there are new versions of these deps, we already have a PR here that updates them (e.g. #487 ). Dependabot is just creating spam by opening PRs when there's already one that would update these deps. Also, because these deps depend on each other, the PRs created automatically are never ready.

PR #523, opened 2023-07-24
Remove `bundled` feature of `rusqlite` for iOS target

andrei-21:
`bundled` enables `modern_sqlite` feature (i.e. it uses the API of recent versions of SQLite). But iOS does not use the bundled version (maybe it doesn't get bundled for iOS) and iOS native version of SQLite is not very recent.

### DIFF
--- a/eel/Cargo.toml
+++ b/eel/Cargo.toml
@@ -28,7 +28,6 @@ num_enum = "0.6.1"
 prost = "0.11.9"
 rand = "0.8.5"
 reqwest = { version = "0.11.18", default-features = false, features = ["blocking", "rustls-tls"] }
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-release.yml
```diff
@@ -105,7 +105,7 @@ jobs:
 
   lipo-ios:
     name: lipo-ios
-    needs: [ cross-compile-ios, cross-compile-android ]
+    needs: [ cross-compile-ios ]
```
danielgranhao, line 108: Why remove it?

danielgranhao (approved):

gcomte (approved):

PR #543, opened 2023-08-02
Collect offer

andrei-21:


eel/src/lib.rs
```diff
@@ -546,6 +547,61 @@ impl LightningNode {
...
+        } else if amount_msat > max {
+            return Err(runtime_error(
+                RuntimeErrorCode::LNURLError,
+                format!("Expected: {amount_msat} msat is above max withdrawable: {max} msat"),
+            ));
+        }
+
+        let description = String::new();
+        // TODO: Put info about the offer.
+        let metadata = "".to_string();
```
gcomte, line 588: What's the difference between `"".to_string();` and `String::new();`? Can't `description` and `metadata` be initiated in the same way?

gcomte (approved):

PR #553, opened 2023-08-09
Expose `ExchangeRate` for Pocket offer rather than `FiatValue`

andrei-21:
It makes more sense since the fiat value was not exchanged *from* sats. Also it will simplify the persistance of offers.

### DIFF
--- a/src/lipalightninglib.udl
+++ b/src/lipalightninglib.udl
@@ -411,8 +411,10 @@ dictionary OfferInfo {
 interface OfferKind {
     Pocket(
         string id,
-        FiatValue topup_value, // The original fiat amount sent to the exchange
-        FiatValue exchange_fee, // The fee paid to perform the exchange from fiat to sats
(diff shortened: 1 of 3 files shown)


src/lipalightninglib.udl
```diff
@@ -410,8 +410,9 @@ dictionary OfferInfo {
 interface OfferKind {
     Pocket(
         string id,
-        FiatValue topup_value, // The original fiat amount sent to the exchange
-        FiatValue exchange_fee, // The fee paid to perform the exchange from fiat to sats
+        ExchangeRate exchange_rate, // The exchange rate used by the exchange to exchange fiat to sats
+        u64 topup_value_minor_units, // The original fiat amount sent to the exchange
+        u64 exchange_fee_minor_units, // The fee paid to perform the exchange from fiat to sats
```
gcomte, line 415: I assume this is in the original Fiat currency as well? This is not explicitly mentioned.

gcomte (approved):

PR #573, opened 2023-08-17
Update transitive dependencies

gcomte:


dleutenegger:
Were you able to check if this may break our iOS build @gcomte ? Such as it did in [!572](https://github.com/getlipa/lipa-lightning-lib/pull/572)

gcomte:
You mean [this one](https://github.com/getlipa/lipa-lightning-lib/actions/runs/5877954902), right?

No, I don't know how to reproduce it. The Android/iOS build test ran through successfully (but so did it the last time when the error occurred).

I'd just try it and in case the error appears again debug the error and find a better fix/check or in the worst case just revert this PR.

dleutenegger (approved):

PR #585, opened 2023-08-25
Add and expose payment uuid

dleutenegger:


src/lib.rs
```diff
@@ -423,6 +424,12 @@ impl LightningNode {
     pub fn get_wallet_pubkey_id(&self) -> Option<String> {
         self.auth.get_wallet_pubkey_id()
     }
+
+    pub fn get_payment_uuid(&self, payment_hash: String) -> String {
+        Uuid::new_v5(&Uuid::NAMESPACE_OID, payment_hash.as_bytes())
```
gcomte, line 429: AFAIK, the payment hash is a HEX-string. What you are doing here with `payment_hash.as_bytes()`, is you take the index of the UTF8 encoding of each character as a binary number.

While this will probably work just fine, to me it feels somewhat wrong. Instead you should turn the hex value into bytes (e.g. `hex::decode()` or something). Then you would have a UUID from the actual hash, not from the UTF-8 encoding of a string of the hash.

This would also save us from confusion if at some point we derive a UUID from a `Hash` rather than a `String`, as then the result will be the same for the same hash.

gcomte (approved):

PR #595, opened 2023-09-01
Implement node_info

dleutenegger:


src/lib.rs
```diff
@@ -264,8 +267,26 @@ impl LightningNode {
...
+            node_pubkey: node_state.id,
+            peers: node_state.connected_peers,
+            channels_info: ChannelsInfo {
+                num_channels: 0,
+                num_usable_channels: 0,
+                local_balance: (node_state.channels_balance_msat + node_state.onchain_balance_msat)
+                    .to_amount_down(&rate),
+                inbound_capacity: node_state.max_receivable_msat.to_amount_down(&rate),
+                outbound_capacity: node_state.max_payable_msat.to_amount_down(&rate),
+                total_channel_capacities: node_state.max_payable_msat.to_amount_down(&rate),
```
danielgranhao, line 287: This value (`total_channel_capacities`) refers to the total channel capacities. I see no way of us getting this value from the SDK's interface, but I'm also not seeing why we would need it. I propose we remove it.

src/lib.rs
```diff
@@ -264,8 +267,26 @@ impl LightningNode {
...
+        let rate = self.get_exchange_rate();
+
+        Ok(NodeInfo {
+            node_pubkey: node_state.id,
+            peers: node_state.connected_peers,
+            channels_info: ChannelsInfo {
+                num_channels: 0,
+                num_usable_channels: 0,
+                local_balance: (node_state.channels_balance_msat + node_state.onchain_balance_msat)
+                    .to_amount_down(&rate),
```
danielgranhao, line 284: I wouldn't put the sum of these 2 balances in here. The name `local_balance` could indeed refer to both balances, but it breaks our existing model/assumptions. We have to think about how to manage the onchain balance, but for now, I would just do the following:
```suggestion
                local_balance: node_state.channels_balance_msat.to_amount_down(&rate),
```
We could already expose the onchain balance as a separate field, leaving to the consumer the decision to display the sum or not.

src/lib.rs
```diff
@@ -264,8 +267,26 @@ impl LightningNode {
...
+            msg: e.to_string(),
+        })?;
+        let rate = self.get_exchange_rate();
+
+        Ok(NodeInfo {
+            node_pubkey: node_state.id,
+            peers: node_state.connected_peers,
+            channels_info: ChannelsInfo {
+                num_channels: 0,
+                num_usable_channels: 0,
```
danielgranhao, line 282: If we keep this here, please add a comment on both lines mentioning the deprecation of the fields, as well as in the UDL.

src/lib.rs
danielgranhao: Hum... I don't quite get what that field means. It seems to be a value about the LSP, not about our current existing channels

dleutenegger: I'm not sure if I understand this correctly, but wouldn't `total_channel_capacities` be the sum of `inbound_capacity` and `outbound_capacity`

danielgranhao: Besides channel reserves, there are other amounts that are not included in the `inbound`/`output` capacity, such as commit fees, so no, unfortunately we can't get the overall channel capacity that way

danielgranhao (approved):

PR #601, opened 2023-09-04
Implement fee protection

dleutenegger:


danielgranhao (approved):

src/lib.rs
```diff
@@ -324,10 +327,10 @@ impl LightningNode {
         todo!()
     }
 
-    // TODO remove unused_variables after breez sdk implementation
-    #[allow(unused_variables)]
-    pub fn get_payment_max_routing_fee_mode(&self, amount_sat: u64) -> MaxRoutingFeeMode {
-        todo!()
+    pub fn get_payment_max_routing_fee_mode(&self, _amount_sat: u64) -> MaxRoutingFeeMode {
+        MaxRoutingFeeMode::Relative {
```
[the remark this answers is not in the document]
dleutenegger: How would you suggest accommodating for this?

As far as I understand, if the fee is smaller than `exemptfee` we ignore the `maxfeepercent`. But in this method we do not know about the fees. As written in the Jira Issue it might make sense to deprecate this method.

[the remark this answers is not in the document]
dleutenegger: Okay you're right. Why deprecating it, well couldn't we just let the node handle this for us or are we exposing this due to UX benefits?

src/lib.rs
```diff
@@ -324,10 +329,16 @@ impl LightningNode {
         todo!()
     }
 
-    // TODO remove unused_variables after breez sdk implementation
-    #[allow(unused_variables)]
     pub fn get_payment_max_routing_fee_mode(&self, amount_sat: u64) -> MaxRoutingFeeMode {
-        todo!()
+        if amount_sat * (MAX_FEE_PERMYRIAD as u64) / 100 < EXEMPT_FEE_MSAT {
```
danielgranhao, line 333: It should be like follows, right? Maybe we can add some unit tests to catch these types of errors.
```suggestion
        if amount_sat * (MAX_FEE_PERMYRIAD as u64) / 10 < EXEMPT_FEE_MSAT {
```

danielgranhao: Another option to make it more readable:
```suggestion
        if amount_sat * ((MAX_FEE_PERMYRIAD as u64) / 10_000) * 1_000 < EXEMPT_FEE_MSAT {
```

danielgranhao: Ah yes, forget my latest comment

danielgranhao (approved):

PR #602, opened 2023-09-04
Implement `decode_invoice()`

danielgranhao:


src/lib.rs
```diff
@@ -321,7 +322,22 @@ impl LightningNode {
         &self,
```
dleutenegger, line 322: Remove `TODO` from function

src/lib.rs
```diff
@@ -321,7 +322,22 @@ impl LightningNode {
...
         invoice: String,
     ) -> std::result::Result<InvoiceDetails, DecodeInvoiceError> {
-        todo!()
+        match self.rt.block_on(parse(&invoice)) {
+            Ok(input_type) => match input_type {
+                InputType::Bolt11 {
```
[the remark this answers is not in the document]
danielgranhao: I've just tried it, and it doesn't help much because we still need to handle the other `InputType`s, meaning that we have to add a catch-all `_ => {}` term anyway.

dleutenegger (approved):

src/invoice_details.rs
```diff
@@ -1,5 +1,7 @@
-use crate::amount::Amount;
+use crate::amount::{Amount, ToAmount};
 
+use crate::ExchangeRate;
+use breez_sdk_core::LNInvoice;
 use std::time::{Duration, SystemTime};
 
 // TODO remove dead code after breez sdk implementation
```
gcomte, line 7: // TODO remove dead code after breez sdk implementation
#[allow(dead_code)]

These 2 lines can be removed, right?

danielgranhao: Done! Ty for noticing

gcomte (approved):

PR #606, opened 2023-09-05
Simplify `decode_invoice()`

andrei-21:
That is what I meat by collapsing two match statements. Is it cleaner now?


danielgranhao (approved):

danielgranhao:
LGTM! I didn't think of the `Ok(_) => ...` so it ended up getting worse but like this it's good. Ty

PR #623, opened 2023-09-06
Add user preferences

andrei-21:


src/lib.rs
```diff
@@ -173,16 +173,21 @@ pub struct OfferInfo {
...
 
+pub(crate) struct UserPreferences {
+    fiat_currency: String,
+    timezone_config: TzConfig,
+}
+
 // TODO remove dead code after breez sdk implementation
 #[allow(dead_code)]
 pub struct LightningNode {
+    user_preferences: Arc<Mutex<UserPreferences>>,
```
gcomte, line 184: Seems to be somewhat redundant with the field `data_store`

danielgranhao (approved):

src/lib.rs
gcomte: You now have the field `user_preferences` that stores data of the type `UserPreferences` and also the field `data_store`, which stores data of the type `UserPreferences` (and `Connection`)

PR #626, opened 2023-09-06
Implement exchange rates

dleutenegger:
Used the fiat currency config from #623


src/data_store.rs
danielgranhao, line 1: Can we add some tests here? You should be able to use the test `test_exchange_rate_storage()` that we had in `eel`

danielgranhao (approved):

PR #628, opened 2023-09-06
Store payment info

andrei-21:


src/data_store.rs
```diff
@@ -70,6 +128,39 @@ impl DataStore {
     }
 }
 
+// Store all provided exchange rates.
+// For every row it takes ~13 bytes (4 + 3 + 2 + 4), if we have 100 fiat currencies it adds 1300 bytes.
+// For 1000 payments it will add ~1 MB.
+fn insert_snapshot(
```
gcomte, line 134: not very idiomatic function name. Isn't it a snapshot of exchange rates? How about `snapshot_exchange_rates`? Or a shorter version would be `snap_rates`. Or any combination of this.

src/lib.rs
```diff
@@ -497,6 +527,24 @@ impl LightningNode {
...
+                        &invoice.payment_hash,
+                        user_preferences,
+                        exchange_rates,
+                        None,
+                    )
+                    .map_to_permanent_failure("Failed to persist payment info")
+            }
+            _ => Err(invalid_input("Invalid invoice")),
+        }?;
+
```
gcomte, line 547: `pay_invoice()` and `pay_open_invoice()` now duplicate a lot of code. We should probably create a private function that contains that code and is being called by `pay_invoice()` and `pay_open_invoice()`. Maybe in another PR.

src/data_store.rs
danielgranhao: Sure. Changed in [a761e80](https://github.com/getlipa/lipa-lightning-lib/pull/628/commits/a761e80dc06674ed86422f8537484ce3d8330912)

src/lib.rs
danielgranhao: There's still some duplicated code, but it got improved in [c90d94e](https://github.com/getlipa/lipa-lightning-lib/pull/628/commits/c90d94ef4ad97535821a2a4f9fb5af7259137126). I would leave more improvements, if any, for later PRs.

gcomte (approved):

PR #629, opened 2023-09-07
LDK node: Upgrade crow & expose offer status REFUNDED

gcomte:
Same as https://github.com/getlipa/lipa-lightning-lib/pull/622, but for the LDK node setup, not Breez SDK

### DIFF
--- a/src/lipalightninglib.udl
+++ b/src/lipalightninglib.udl
@@ -152,7 +152,7 @@ interface LightningNode {
 
     // Get a list of currently available fund offers
     [Throws=LnError]
-    sequence<OfferInfo> query_available_offers();
+    sequence<OfferInfo> query_uncompleted_offers();
 
     // Request to collect the offer.
     // A payment hash will be returned to track incoming payment.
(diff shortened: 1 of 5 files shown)


src/lipalightninglib.udl
```diff
@@ -437,7 +437,8 @@ interface OfferKind {
 
 enum OfferStatus {
     "READY",
-    "FAILED",
+    "FAILED", // Claiming the offer failed, but it can be retried.
+    "REFUNDED", // The offer was not claimed in time, so the user got refunded (Pocket: Refund happened over the Fiat rails).
```
dleutenegger, line 441: That the offer was not claimed in time is only one of a few reasons why an offer got refunded. A full list of reasons for Pocket can be found [here](https://pocketbitcoin.com/developers/docs/rest/v1/webhooks#refund-reasons).

gcomte: Awesome, thanks!

danielgranhao (approved):

gcomte:
Fixed the docs in the UDL file @dleutenegger

dleutenegger (approved):

PR #635, opened 2023-09-08
Implement LSP fee caching

dleutenegger:


src/task_manager.rs
```diff
@@ -47,6 +58,24 @@ impl TaskManager {
...
+            None => Err(RuntimeError {
+                code: RuntimeErrorCode::LspServiceUnavailable,
+                msg: "Cached lsp fee isn't available".to_string(),
+            }),
+        }?;
+
+        Ok(LspFee {
+            channel_minimum_fee: lsp_fee_cache
+                .channel_minimum_fee
+                .to_amount_down(exchange_rate),
```
danielgranhao, line 74: We should be pessimistic from the user's POV. As this is a fee, it's better to round up.
```suggestion
                .to_amount_up(exchange_rate),
```

src/task_manager.rs
```diff
@@ -20,6 +24,7 @@ pub(crate) struct TaskManager {
     exchange_rates: Arc<Mutex<Vec<ExchangeRate>>>,
     data_store: Arc<Mutex<DataStore>>,
     sdk: Arc<BreezServices>,
+    lsp_fee: Arc<Mutex<Option<LspFee>>>,
```
danielgranhao, line 27: Instead of storing an `LspFee`, I would store the `OpeningFeeParams` provided by Breez. That way, we avoid having to use the following expressions, which are a bit weird :) 
`cheapest_opening_fee.min_msat.to_amount_up(&None)`
and 
`(lsp_fee_cache.channel_minimum_fee.sats * 1000)
                .to_amount_up(exchange_rate)`
This way, you just have to convert from `msat` to `Amount` in `get_lsp_fee()`

```suggestion
    lsp_fee: Arc<Mutex<Option<OpeningFeeParams>>>,
```

PR #637, opened 2023-09-08
Implement payment info retrieval

danielgranhao:


src/lib.rs
```diff
@@ -527,16 +533,137 @@ impl LightningNode {
...
+        &self,
+        breez_payment: breez_sdk_core::Payment,
+    ) -> Result<Payment> {
+        let payment_details = if let PaymentDetails::Ln { data } = breez_payment.details {
+            data
+        } else {
+            return Err(permanent_failure(
+                "Current interface doesn't support PaymentDetails::ClosedChannel",
+            ));
+        };
```
gcomte, line 583:
```suggestion
        match breez_payment.details {
            PaymentDetails::Ln { data } => data,
            _ => return Err(permanent_failure("Current interface doesn't support PaymentDetails::ClosedChannel")),
        }
```

gcomte (approved):

src/lib.rs
danielgranhao: Applied in [ce72943](https://github.com/getlipa/lipa-lightning-lib/pull/637/commits/ce7294377591216ade8018100c4a24b3b1118322)

dleutenegger (approved):

PR #640, opened 2023-09-11
Update time dep + remove audit ignore

danielgranhao:


Cargo.lock
```diff
@@ -702,16 +702,15 @@ dependencies = [
 
 [[package]]
 name = "chrono"
-version = "0.4.29"
+version = "0.4.30"
```
dleutenegger, line 705: Would probably make sense to update the version for this dependency in the `Cargo.toml`

dleutenegger (changes requested):

Cargo.lock
danielgranhao: Oh, didn't realize we also depend on `chrono` ourselves. Updated in [c4f45f4](https://github.com/getlipa/lipa-lightning-lib/pull/640/commits/c4f45f41f1a2f6ad4c59c6d08a2b2e6d092edecc).

dleutenegger (approved):

PR #641, opened 2023-09-11
Add `libc++_shared.so` to jniLibs

danielgranhao:


.github/workflows/deploy-release.yml
```diff
@@ -296,6 +296,13 @@ jobs:
           mkdir -p jniLibs/arm64-v8a
           mkdir -p jniLibs/armeabi-v7a
           mkdir -p jniLibs/x86
+          
+          wget https://dl.google.com/android/repository/android-ndk-r25c-windows.zip
+          unzip android-ndk-r25c-windows.zip
+          cp android-ndk-r25c/toolchains/llvm/prebuilt/windows-x86_64/sysroot/usr/lib/aarch64-linux-android/libc++_shared.so jniLibs/arm64-v8a
+          cp android-ndk-r25c/toolchains/llvm/prebuilt/windows-x86_64/sysroot/usr/lib/arm-linux-androideabi/libc++_shared.so jniLibs/armeabi-v7a
+          cp android-ndk-r25c/toolchains/llvm/prebuilt/windows-x86_64/sysroot/usr/lib/i686-linux-android/libc++_shared.so jniLibs/x86
```
gcomte, line 304: Is there a specific reason for why you don't `mv` instead? That should be faster.

```suggestion
          mv android-ndk-r25c/toolchains/llvm/prebuilt/windows-x86_64/sysroot/usr/lib/aarch64-linux-android/libc++_shared.so jniLibs/arm64-v8a
          mv android-ndk-r25c/toolchains/llvm/prebuilt/windows-x86_64/sysroot/usr/lib/arm-linux-androideabi/libc++_shared.so jniLibs/armeabi-v7a
          mv android-ndk-r25c/toolchains/llvm/prebuilt/windows-x86_64/sysroot/usr/lib/i686-linux-android/libc++_shared.so jniLibs/x86
```

gcomte (approved):

.github/workflows/deploy-release.yml
danielgranhao: No reason. I'm accepting your suggestion. But keep in mind these files are <10MB so any improvement in speed should be insignificant :)

gcomte (approved):

PR #643, opened 2023-09-12
Always use SDK prod

danielgranhao:
Given that the Breez SDK staging env isn't really usable (no lsp available), let't always use prod.


danielgranhao:
Nope, we'll remove them, this was just in the spirit of not changing the interface yet.

PR #644, opened 2023-09-13
Persist funds migration status

andrei-21:


src/fund_migration.rs
```diff
@@ -5,8 +5,19 @@ use bitcoin::hashes::sha256;
...
 use bitcoin::util::bip32::{ChildNumber, ExtendedPrivKey};
 use bitcoin::Network;
+use num_enum::TryFromPrimitive;
 use perro::MapToError;
 
+#[derive(PartialEq, Eq, Debug, TryFromPrimitive, Clone)]
+#[repr(u8)]
+pub enum MigrationStatus {
+    Unknown,
+    Pending,
```
danielgranhao, line 15: Do we need a `Pending` status? We don't need to wait for the invoice to be paid as the backend will do that for us before returning successfully, so I don't see a need for it.

gcomte (approved):

PR #646, opened 2023-09-13
Limit amount of payments being listed

gcomte:
This is a temporary solution optimized to fit the current 3L interface.

It is not optimized for the interaction with the Breez SDK interface. This is meant to happen in a future PR.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -538,26 +538,17 @@ impl LightningNode {
 
     // TODO: restrict returned payments according to `number_of_payments` parameter
     pub fn get_latest_payments(&self, _number_of_payments: u32) -> Result<Vec<Payment>> {
-        let breez_payments = self
-            .rt
+        self.rt
             .handle()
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -536,28 +536,19 @@ impl LightningNode {
...
-    pub fn get_latest_payments(&self, _number_of_payments: u32) -> Result<Vec<Payment>> {
-        let breez_payments = self
-            .rt
+    pub fn get_latest_payments(&self, number_of_payments: u32) -> Result<Vec<Payment>> {
+        self.rt
             .handle()
             .block_on(self.sdk.list_payments(PaymentTypeFilter::All, None, None))
             .map_to_runtime_error(
                 RuntimeErrorCode::NodeUnavailable,
                 "Failed to get payment by hash",
```
danielgranhao, line 545: This problem was already here, but since you're already here, could you fix the error message? :) 

```suggestion
                "Failed to list payments",
```

danielgranhao (approved):

PR #647, opened 2023-09-13
Remove unused errors

danielgranhao:
This presumes that we are fine with starting to change the interface


danielgranhao:
@andrei-21 Not sure that makes sense. We are blocked from doing the migration until Breez SDK is ready to onboard users.

PR #648, opened 2023-09-14
Migrate funds

andrei-21:


src/fund_migration.rs
```diff
@@ -49,11 +156,50 @@ mod tests {
...
+            valid_until: String::new(),
+            max_idle_time: 0,
+            max_client_to_self_delay: 0,
+            promise: String::new(),
+        };
+
+		assert_eq!(add_lsp_fees(      1_000, &lsp_fee),   2_001_000);
+		assert_eq!(add_lsp_fees(  1_000_000, &lsp_fee),   3_000_000);
+		assert_eq!(add_lsp_fees( 11_000_000, &lsp_fee),  13_000_000);
+		assert_eq!(add_lsp_fees(811_000_000, &lsp_fee), 819_191_000);
```
gcomte, line 182: According to my calcualtion, this should be:

10_000 ppm = 1%
811_000_000 * 1.01 = 819_110_000

src/fund_migration.rs
```diff
@@ -19,25 +24,127 @@ pub enum MigrationStatus {
...
+    }
+
+    let one_minus_proportional = MILLION - lsp_fee.proportional as u64;
+
+    if amount_msat * lsp_fee.proportional as u64 / one_minus_proportional < lsp_fee.min_msat {
+        amount_msat + lsp_fee.min_msat
+    } else {
+        let result = amount_msat * MILLION / one_minus_proportional;
+        result / 1_000 * 1_000
+    }
```
gcomte, line 124: Not sure what you're calculating here. I'm coming up with a different calculation which also results in different amounts :thinking: 

Tell me what I'm missing here:

```suggestion
    let proportional_fee = amount_msat * lsp_fee.proportional as u64 / MILLION;

    if proportional_fee < lsp_fee.min_msat {
        amount_msat + lsp_fee.min_msat
    } else {
        let result = amount_msat + proportional_fee;
        result / 1_000 * 1_000
    }
```

src/fund_migration.rs
```diff
@@ -19,25 +24,127 @@ pub enum MigrationStatus {
...
+    let public_key = public_key.serialize().to_hex();
+
+    let balance = fetch_balance(public_key.clone())?;
+    if balance == 0 {
+        data_store
+            .lock()
+            .unwrap()
+            .append_funds_migration_status(MigrationStatus::NotNeeded)?;
+        return Ok(());
+    }
```
danielgranhao, line 52: From my understanding of the migration service, it will return errors (distinct ones) if the node never had any funds to migrate or if it already migrated. This snippet assumes it would return 0, at least in the former case.

src/fund_migration.rs
```diff
@@ -19,25 +24,127 @@ pub enum MigrationStatus {
...
+
+    let lsp_info = sdk.lsp_info().await.map_to_runtime_error(
+        RuntimeErrorCode::LspServiceUnavailable,
+        "Failed to get LSP info",
+    )?;
+    let lsp_fee = lsp_info
+        .opening_fee_params_list
+        .get_cheapest_opening_fee_params()
+        .map_to_permanent_failure("Failed to get LSP fees")?;
+    let amout_to_request = add_lsp_fees(balance, &lsp_fee) * 1_000;
```
danielgranhao, line 67: We don't need to implement this. The SDK provides `BreezServices::open_channel_fee()`, right?

src/fund_migration.rs
```diff
@@ -19,25 +24,127 @@ pub enum MigrationStatus {
...
+
+    let invoice = sdk
+        .receive_payment(breez_sdk_core::ReceivePaymentRequest {
+            amount_sats: amout_to_request,
```
danielgranhao, line 71:
```suggestion
            amount_sats: amount_to_request,
```

src/fund_migration.rs
danielgranhao: Forget this, I forgot the lsp fees are calculated on the requested value 🤦

src/fund_migration.rs
gcomte: As discussed. My assumption here is wrong, because this calculation assumes a payment where the _sent_ amount is an exactly defined number, but what we need for the migration is a payment where the _received_ amount is achieved accurately.

src/fund_migration.rs
gcomte: As discussed. My assumption here is wrong, because this calculation assumes a payment where the _sent_ amount is an exactly defined number, but what we need for the migration is a payment where the _received_ amount is achieved accurately.

src/fund_migration.rs
gcomte: As discussed. The assumption here is wrong, because this calculation assumes a payment where the _sent_ amount is an exactly defined number, but what we need for the migration is a payment where the _received_ amount is achieved accurately.

danielgranhao (approved):

PR #650, opened 2023-09-15
Implement funds migration

andrei-21:
@dleutenegger I noticed that the code will be simpler if `migration_balance()` returns 0 if there is no balance, rather than throwing an error, can we change the backend for that?

**TODO**: Change `graphql` dependency to a new tag once https://github.com/getlipa/wild/pull/65 is merged.

### DIFF
--- a/src/fund_migration.rs
+++ b/src/fund_migration.rs
@@ -1,6 +1,7 @@
 use crate::data_store::DataStore;
 use crate::errors::{Result, RuntimeErrorCode};
 
+use crate::async_runtime::Handle;
 use bitcoin::hashes::hex::ToHex;
 use bitcoin::hashes::sha256;
 use bitcoin::secp256k1::{Message, PublicKey, SecretKey, SECP256K1};
@@ -28,7 +29,8 @@ pub(crate) enum MigrationStatus {
 }
 
 #[allow(dead_code)]
-pub(crate) async fn migrate_funds(
+pub(crate) fn migrate_funds(
+    rt: Handle,
     seed: &[u8; 64],
     data_store: Arc<Mutex<DataStore>>,
     sdk: &BreezServices,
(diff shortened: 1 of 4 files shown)


src/fund_migration.rs
```diff
@@ -96,14 +110,37 @@ pub(crate) async fn migrate_funds(
     }
 }
 
-fn payout(_public_key: String, _invoice: String, _signature: String) -> Result<()> {
-    // TODO: Implement.
-    Ok(())
+fn fetch_legacy_balance(client: &Client, backend_url: &String, public_key: String) -> Result<u64> {
+    let variables = migration_balance::Variables {
+        node_pub_key: Some(public_key),
```
dleutenegger, line 115: Missing the hex string prefix (`\\x`)
```suggestion
        node_pub_key: Some(format!("\\xpublic_key")),
```

src/lib.rs
```diff
@@ -339,8 +338,15 @@ impl LightningNode {
...
-        //     .block_on(async { migrate_funds(&strong_typed_seed, data_store_clone, &sdk).await })?;
+        // let auth_clone = Arc::clone(&auth);
+        // fund_migration::migrate_funds(
+        //     rt.handle(),
+        //     &strong_typed_seed,
+        //     data_store_clone,
+        //     &sdk,
+        //     auth_clone,
+        //     &environment.backend_url,
+        // )?;
```
dleutenegger, line 349: Maybe add a comment as for why we comment this out

dleutenegger (changes requested):

src/fund_migration.rs
dleutenegger: This can actually be ignored. Just went over the service, we do not fully comply with hasuras hex encoding. Lets not send any prefixed hex strings for the migration

danielgranhao: Removed all prefixes in [3e2bf28](https://github.com/getlipa/lipa-lightning-lib/pull/650/commits/3e2bf286a007f49c9f637531a48c6b803f330262)

src/lib.rs
danielgranhao: Added todo comment in [3e2bf28](https://github.com/getlipa/lipa-lightning-lib/pull/650/commits/3e2bf286a007f49c9f637531a48c6b803f330262)

dleutenegger (approved):

PR #671, opened 2023-10-02
Make `decode_invoice()` to return more specific errors

andrei-21:


src/lib.rs
```diff
@@ -391,10 +391,10 @@ impl LightningNode {
...
 
     pub fn get_node_info(&self) -> Result<NodeInfo> {
-        let node_state: NodeState = self.sdk.node_info().map_err(|e| RuntimeError {
-            code: RuntimeErrorCode::NodeUnavailable,
-            msg: e.to_string(),
-        })?;
+        let node_state = self
+            .sdk
+            .node_info()
+            .map_to_permanent_failure("Failed to read node info")?;
```
danielgranhao, line 397: Why permanent failure? The SDK seems to return `ServiceErrorCode::NotReady` in the version that is being worked on. This may not indicate a permanent problem but a runtime one instead.

src/lib.rs
```diff
@@ -491,15 +491,41 @@ impl LightningNode {
...
                 msg: "Failed to decode invoice - provided string was recognized but not as a Bolt11 invoice".to_string(),
             }),
             Err(e) => Err(DecodeInvoiceError::ParseError {
                 msg: format!("Failed to parse invoice: {e}"),
             }),
+        }?;
+        let node_id = self
+            .sdk
+            .node_info()
+            .map_err(|e| DecodeInvoiceError::PermanentFailure {
```
danielgranhao, line 508: Same here, I'm not sure if this is a permanent failure

src/lipalightninglib.udl
```diff
@@ -538,9 +538,12 @@ interface MnemonicError {
 
 [Error]
 interface DecodeInvoiceError {
+    PermanentFailure(string msg);
     ParseError(string msg);
     SemanticError(string msg);
     NetworkMismatch(Network expected, Network found);
+    PayingToSelf();
+    InvoiceExpired();
```
danielgranhao, line 546: I'm not sure we want to fail decoding if the invoice is expired. If I try to pay an expired invoice, I could still be interested in checking out the amount that the payee requested. Indeed this a detail, so if you think it isn't worth it to account for this potential use case, feel free to ignore this.

src/lib.rs
danielgranhao: Maybe if it isn't persisted, it means it will be persisted soon, so still not a permanent failure. I would simply keep this as the `NodeUnavailable` runtime error. 

I don't think we should be looking into implementation of dependencies to figure out what errors to throw. That can easily lead to issues because the dependency implementation can change.

tests/invoice_test.rs
```diff
@@ -0,0 +1,41 @@
...
+    // let result = node.decode_invoice(testnet_invoice);
+    // assert!(matches!(
+    //     result,
+    //     Err(DecodeInvoiceError::NetworkMismatch {
+    //         expected: Network::Bitcoin,
+    //         found: Network::Testnet
+    //     })
+    // ));
+
+    let bitcoi_address = "1DTHjgRiPnCYhgy7PcKxEEWAyFi4VoJpqi".to_string();
```
danielgranhao, line 31:
```suggestion
    let bitcoin_address = "1DTHjgRiPnCYhgy7PcKxEEWAyFi4VoJpqi".to_string();
```

danielgranhao (approved):

PR #685, opened 2023-10-11
Add topup integration tests

dleutenegger:


danielgranhao (commented):
I would appreciate some comments documenting why these offers would be created right after registration. Otherwise, LGTM!

danielgranhao (approved):

raweber42 (approved):

PR #712, opened 2023-10-24
Implement `prepare_refund_failed_swap()`

danielgranhao:


src/lib.rs
```diff
@@ -278,6 +278,14 @@ pub struct FailedSwapInfo {
     pub created_at: SystemTime,
 }
 
+/// Information about a failed swap refund
+pub struct FailedSwapRefundInfo {
```
dleutenegger, line 282: Maybe rename this to `RefundFailedSwapInfo`, otherwise it's a bit confusing if a swap refund has failed or if a swap has failed and this is about refunding.

src/lipalightninglib.udl
```diff
@@ -85,6 +85,9 @@ interface LightningNode {
     [Throws=LnError]
     sequence<FailedSwapInfo> get_unresolved_failed_swaps();
 
+    [Throws=LnError]
+    FailedSwapRefundInfo prepare_refund_failed_swap(FailedSwapInfo failed_swap_info, string to_address, u32 onchain_fee_rate);
```
dleutenegger, line 89: Might make sense to expand the `FailedSwapRefundInfo` to be used as an argument in the `resolve_failed_swap` method.

dleutenegger (changes requested):

src/lipalightninglib.udl
danielgranhao: I think we don't need to force the use of the prepare method. Why do we need more info other than the swap address?

src/lib.rs
danielgranhao: Addressed in [b7788fc](https://github.com/getlipa/lipa-lightning-lib/pull/712/commits/b7788fc89a1140e66119e0c281e8f24ba8eb83c0)

src/lipalightninglib.udl
dleutenegger: I was just thinking from the client perspective, but I might have misunderstood the use of the `prepare_refund_failed_swap` method.

As a client if, and here I'm not sure if that's the case, I always call the `prepare_refund_failed_swap` method before refunding a swap would prefer to just supply the response from this to the `refund_failed_swap` method.
But if the client doesn't always call the prepare method before, this doesn't make sense.

dleutenegger: I was just thinking from the client perspective, but I might have misunderstood the use of the prepare_refund_failed_swap method.

As a client if, and here I'm not sure if that's the case, I always call the prepare_refund_failed_swap method before refunding a swap would prefer to just supply the response from this to the refund_failed_swap method.
But if the client doesn't always call the prepare method before, this doesn't make sense.

dleutenegger (approved):

PR #727, opened 2023-11-03
Update errors

andrei-21:


src/lib.rs
```diff
@@ -728,24 +733,8 @@ impl LightningNode {
...
-            })) {
-            Ok(_) => Ok(()),
-            // TODO: properly handle errors (requires changing either ours or the SDK's error model)
-            Err(e) => Err(RuntimeError {
-                code: PayErrorCode::UnexpectedError,
-                msg: format!("Failed to start paying invoice: {e}"),
-            }),
-        }
+    pub fn pay_invoice(&self, invoice_details: InvoiceDetails, metadata: String) -> PayResult<()> {
+        self.pay_open_invoice(invoice_details, 0, metadata)
```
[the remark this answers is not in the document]
dleutenegger: My bad, I meant that we eventually pass `None` in `breez_services::send_payment` for the amount, but if `0` works as well, I'm fine with it. No need to unnecessarly adjust the interface.

src/errors.rs
```diff
@@ -29,33 +29,40 @@ pub type Result<T> = std::result::Result<T, Error>;
...
-    /// Retry attempts or timeout was reached.
-    /// It might make sense to retry the payment.
-    RetriesExhausted,
-    /// All possible routes failed.
+
+    /// A locally issued invoice tried to be paid. Self-payments are not supported.
+    /// There's no point in retrying this payment.
+    PayingToSelf,
+
+    /// The payment failed. Likely the issue is on the receiver.
```
danielgranhao, line 50: Are we sure about this?

src/errors.rs
```diff
@@ -29,33 +29,40 @@ pub type Result<T> = std::result::Result<T, Error>;
...
+    /// Payment timed out.
     /// It might make sense to retry the payment.
-    NoMoreRoutes,
-    /// An unexpected error occurred. This likely is a result of a bug within 3L/LDK and should be reported to lipa.
-    ///
-    /// *WARNING* At the moment, all payment failures will return this code. Once Breez SDK reworks their error model, we'll
-    /// be able to provide much more specific error codes, such as the other ones that are part of this enum.
+    PaymentTimeout,
+
+    /// Route too expensive. The fee the route exceeds the settings.
```
danielgranhao, line 58:
```suggestion
    /// Route too expensive. The route's fee exceeds the settings.
```

src/errors.rs
```diff
@@ -68,6 +75,42 @@ impl Display for PayErrorCode {
...
+    NoRouteFound,
+
+    /// The payment failed. Likely the issue is on the receiver.
+    PaymentFailed,
+
+    /// Payment timed out.
+    /// It might make sense to retry the payment.
+    PaymentTimeout,
+
+    /// Route too expensive. The fee the route exceeds the settings.
```
danielgranhao, line 94:
```suggestion
    /// Route too expensive. The route's fee exceeds the settings.
```

src/lib.rs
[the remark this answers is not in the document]
danielgranhao: @andrei-21 We should at least in `pay_open_invoice` provide `None` to breez when `amount_sat` is `0`. Even if it's working now, Breez can change the implementation in the future.

src/errors.rs
danielgranhao: Yes... The docs need to be improved. My interpretation of the code name is that this is a generic error, and not just for when the issue is the receiver. Some of these CLN errors seems like they could indicate a bug with our side (not 3L, but the payer implementation in general)

danielgranhao (approved):

PR #749, opened 2023-11-10
Add analytics service

dleutenegger:
- [ ] Update wild versions


src/analytics.rs
```diff
@@ -0,0 +1,194 @@
...
+        let paid_amount_msat = match paid_amount {
+            None => invoice_amount.unwrap(),
+            Some(a) => a,
+        };
+
+        let user_currency = self.user_preferences.lock_unwrap().fiat_currency.clone();
+        let analytics_client = self.analytics_client.clone();
+
+        self.rt_handle.spawn(async move {
+            log_error(
```
danielgranhao, line 67: This is a suggestion to make it clear that we are ignoring the error. Feel free to ignore it if you don't like it.
```suggestion
            log_ignore_error(
```

danielgranhao: Another improvement would be to implement a method on `Result` such that we call `analytics_client
                    .report_event(...).await.log_ignore_error()`

danielgranhao (approved):

PR #750, opened 2023-11-10
Expose SwapInfo for payments

dleutenegger:


src/lib.rs
```diff
@@ -1015,6 +1025,17 @@ impl LightningNode {
...
+        let user_preferences = self.user_preferences.lock_unwrap();
+        let swap = payment_details.swap_info.map(|s| SwapInfo {
+            bitcoin_address: s.bitcoin_address,
+            created_at: TzTime {
+                time: unix_timestamp_to_system_time(s.created_at as u64),
+                timezone_id: user_preferences.timezone_config.timezone_id.clone(),
+                timezone_utc_offset_secs: user_preferences.timezone_config.timezone_utc_offset_secs,
+            },
+            paid_sats: s.paid_sats,
+        });
```
danielgranhao, line 1037: For the payments resulting from swaps, we won't have local data in the db, right? We should think about how to address that. Even if, in the meantime, we go with this approach of using the system settings, this isn't congruent with the rest of the implementation, and we should add a todo here.
It's possible that if [this](https://github.com/breez/breez-sdk/issues/518) is implemented, we can create payment data in our db when the LN payment is actually made.

danielgranhao (approved):

PR #764, opened 2023-11-17
Expose reset topup functionality

dleutenegger:


src/data_store.rs
```diff
@@ -289,6 +289,14 @@ impl DataStore {
         Ok(())
     }
 
+    pub fn clear_fiat_topup_info(&self) -> Result<()> {
+        self.conn
+            .execute("DELETE FROM fiat_topup_info;", params![])
+            .map_to_permanent_failure("Failed to delete fiat topup info")?;
+
+        Ok(())
+    }
```
danielgranhao, line 298: Just opening it for discussion: could we want information about previous topups later? If so, we could disable a topup info instead of deleting it. One could also argue that we can implement that later if needed, of course.

dleutenegger: Right now, I wouldn't see any reason for keeping old topup data. Where would you see that become handy?

danielgranhao: A user might want to have multiple exchange setups using different bank accounts. This is a likely use case, but let's not try to implement things before we need them. We can easily change this later.

danielgranhao (approved):

PR #769, opened 2023-11-17
Extract payment logic into own file

dleutenegger:


danielgranhao:
I would prefer to keep all `LightningNode` methods in the same file even it if results in a bigger file.

PR #785, opened 2023-11-21
Persist LSP channel opening fees for created invoices

dleutenegger:


src/lib.rs
```diff
@@ -933,25 +935,31 @@ impl LightningNode {
...
+            .map(|f| f.as_msats().to_amount_up(&exchange_rate));
+        let invoice_amount = invoice_details
+            .amount
+            .clone()
+            .ok_or_permanent_failure("Locally created invoice doesn't include an amount")?
+            .sats;
+
+        let amount = (invoice_amount - lsp_fees.clone().map_or(0, |f| f.sats))
+            .as_sats()
+            .to_amount_down(&exchange_rate);
```
danielgranhao, line 949: The `amount` in `Payment` is supposed to be the amount specified in the invoice, not necessarily the amount the user received/will receive.

dleutenegger: Addressed in 1fc2644

danielgranhao (approved):

PR #786, opened 2023-11-21
Fix duplicate funding payment in payment list

dleutenegger:


src/lib.rs
```diff
@@ -719,15 +719,15 @@ impl LightningNode {
 
         let breez_payment_invoices: HashSet<String> = breez_payments
```
danielgranhao, line 720:
```suggestion
        let breez_payment_hashes: HashSet<String> = breez_payments
```

danielgranhao (approved):

src/lib.rs
dleutenegger: Addressed in af4a20c

danielgranhao (approved):

PR #802, opened 2023-11-30
Fix initiating a new CLI sample node

gcomte:


gcomte:
Is likely being fixed by https://github.com/getlipa/lipa-lightning-lib/pull/825

PR #821, opened 2023-12-12
Add `overview` command for the cli app

andrei-21:


gcomte (approved):
It's hacky, but its helpful and it is extracted out into `overview.rs` for the most part

PR #822, opened 2023-12-12
Log 3L version

andrei-21:


gcomte (commented):
Does this just print the branch name?

https://github.blog/changelog/2023-09-13-github-actions-updates-to-github_ref-and-github-ref/

dleutenegger:
Would prefer if we also log this on startup, as it allows seeing when a version change would have happened in the log history

dleutenegger (approved):

PR #855, opened 2024-01-04
Implement invoice affordability functionality

dleutenegger:


src/lib.rs
```diff
@@ -603,6 +615,45 @@ impl LightningNode {
         get_payment_max_routing_fee_mode(amount_sat, &self.get_exchange_rate())
     }
 
+    /// Get the affordability status of an invoice. Throws an error if the invoice doesn't have a
+    /// requested amount.
```
danielgranhao, line 619: Should we change the method to also allow affordability checks for open invoices? Either allow an optional amount to also be provided, or even change the method to only take a sat amount (seems like we only care about the amount in the invoice).

danielgranhao: If we take only a sat amount, it would be useful not only for paying bolt11 invoices, but also LNURL and LN addresses.

src/lib.rs
```diff
@@ -603,6 +615,39 @@ impl LightningNode {
...
+    /// Parameters:
+    /// * `amount` - The to be spent amount.
+    pub fn get_invoice_affordability(&self, amount_sat: u64) -> Result<InvoiceAffordability> {
+        let amount = amount_sat.as_sats();
+
+        let routing_fee_mode = self.get_payment_max_routing_fee_mode(amount_sat);
+
+        let max_fee_msats = match routing_fee_mode {
+            MaxRoutingFeeMode::Relative { max_fee_permyriad } => {
+                (amount_sat / 10000 * (max_fee_permyriad as u64)).as_sats()
```
danielgranhao, line 629:
```suggestion
                (amount_sat * (max_fee_permyriad as u64) / 10000).as_sats()
```

danielgranhao (approved):

PR #859, opened 2024-01-05
Add filter and log for corrupted payments

dleutenegger:


danielgranhao (approved):
LGTM. I would just add a TODO comment to make it clear we want to remove this eventually.

PR #860, opened 2024-01-05
Implement clear wallet

danielgranhao:
Implements preparation and start of a reverse swap.

### DIFF
--- a/examples/node/cli.rs
+++ b/examples/node/cli.rs
@@ -652,7 +652,7 @@ fn print_lnurl_withdraw_details(lnurl_withdraw_details: LnUrlWithdrawDetails) {
 
 fn print_bitcoin_address_data(bitcoin_address_data: BitcoinAddressData) {
     println!("Bitcoin Address data:");
-    println!("  Adress                {}", bitcoin_address_data.address);
+    println!("  Address               {}", bitcoin_address_data.address);
     println!("  Network               {}", bitcoin_address_data.network);
     println!(
         "  Amount SAT            {:?}",
--- a/examples/node/cli.rs
+++ b/examples/node/cli.rs
@@ -2,6 +2,7 @@ use crate::hinter::{CommandHint, CommandHinter};
 use crate::overview::overview;
 
 use anyhow::{anyhow, bail, Context, Result};
+use breez_sdk_core::BitcoinAddressData;
 use chrono::offset::FixedOffset;
 use chrono::{DateTime, Local, Utc};
 use colored::Colorize;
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -1740,6 +1755,73 @@ impl LightningNode {
...
+            clear_amount: amount_sat.as_sats().to_amount_up(&exchange_rate),
+            total_estimated_fees: total_fees_sat.as_sats().to_amount_up(&exchange_rate),
+            onchain_fee: onchain_fee_sat.as_sats().to_amount_up(&exchange_rate),
+            swap_fee: swap_fee_sat.as_sats().to_amount_up(&exchange_rate),
+            fees_hash: reverse_swap_fees.fees_hash,
+        })
+    }
+
+    /// Starts a reverse swap that sends all funds in LN channels to the provided onchain address.
+    pub fn clear_wallet(
```
dleutenegger, line 1805: Mobile might be interested in an identifier of some sorts to navigate to the status page

src/lib.rs
```diff
@@ -1740,6 +1759,112 @@ impl LightningNode {
             .status)
     }
 
+    /// Check if clearing the wallet is feasible.
+    ///
+    /// If not feasible, it means the balance is either too high or too low for a reverse-swap to
+    /// be used.
+    pub fn is_clear_wallet_feasible(&self) -> Result<bool> {
```
gcomte, line 1766: We could be more specific here and return `Result<ClearWalletSate>` or something such that we can actually tell whether the amount was too big or too small.

Once we have that separation, it would be nice to actually be informed about the min / max amount allowed. (I see this could be tricky because of the fees only being an estimation)

danielgranhao: It would be nice to provide that info to the user. I went with the current option because it's enough to support the implementation of the currently planned UI. Even if we change the UI later, should we already implement something that won't be initially used?

gcomte: Well, you're right, _premature optimization is the root of all evil_. However I'm surprised that we don't wont to expose this information in the planned UI.

examples/node/cli.rs
```diff
@@ -632,6 +650,18 @@ fn print_lnurl_withdraw_details(lnurl_withdraw_details: LnUrlWithdrawDetails) {
     );
 }
 
+fn print_bitcoin_address_data(bitcoin_address_data: BitcoinAddressData) {
+    println!("Bitcoin Address data:");
+    println!("  Adress                {}", bitcoin_address_data.address);
```
dleutenegger, line 655:
```suggestion
    println!("  Address               {}", bitcoin_address_data.address);
```

dleutenegger (approved):

PR #880, opened 2024-01-15
Implement recipient node decoding

andrei-21:
image


danielgranhao:
Cool! I know it's a draft, but is the final intention to expose the `decode_recipient` method? I would not expose it and just fill the recipient in `Activities` and paying preparation methods.

gcomte:
In the end it would make sense to store the data in the back-end, to be able to deal with changes quickly. Maybe cache it locally.

PR #913, opened 2024-02-05
Implement analytics opt out

danielgranhao:


src/analytics.rs
```diff
@@ -9,15 +9,24 @@ use breez_sdk_core::{
...
 use std::time::SystemTime;
 use uuid::Uuid;
 
+#[derive(Debug, PartialEq, TryFromPrimitive, Clone)]
+#[repr(u8)]
+pub enum AnalyticsConfig {
+    Enabled,
+    Disabled,
+}
+
```
gcomte, line 24: If this can ever only be `Enabled` or `Disabled`, why not just use a Boolean instead? Are you expecting more sophisticated configurations in the future?

danielgranhao: I opted for the enum approach because it leaves less room for confusion when compared to a bool and also because it allows for more configurations if we ever need them.

PR #917, opened 2024-02-06
Implement unified action required list method

danielgranhao:


examples/node/cli.rs
```diff
@@ -941,57 +958,90 @@ fn list_offers(node: &LightningNode) -> Result<()> {
...
-                    exchange_rate.currency_code,
-                );
-                let exchanged_at: DateTime<Utc> = exchange_rate.updated_at.into();
+    Ok(())
+}
+
+fn print_offer(offer: &OfferInfo) {
+    let kind = match offer.offer_kind {
+        OfferKind::Pocket { .. } => "Pocket",
+    };
```
gcomte, line 971: Should we implement `Display` for `Offerkind` instead? Or is it that you don't want to bloat the Core library with implementations that are only being used in the CLI tool..?

examples/node/cli.rs
```diff
@@ -941,57 +958,90 @@ fn list_offers(node: &LightningNode) -> Result<()> {
...
+                *exchange_fee_rate_permyriad as f64 / 100_f64
+            );
+            println!(
+                "      Exchange fee value: {:.2} {}",
+                *exchange_fee_minor_units as f64 / 100f64,
+                exchange_rate.currency_code,
+            );
+            let exchanged_at: DateTime<Utc> = exchange_rate.updated_at.into();
+            println!(
+                "             Exchange at: {}",
```
gcomte, line 1010: Probably `exchanged at:`, but who cares. This is only the CLI tool.

src/lib.rs
```diff
@@ -276,6 +276,25 @@ pub struct ChannelCloseResolvingFees {
     pub sat_per_vbyte: u32,
 }
 
+#[allow(clippy::large_enum_variant)]
+pub enum ActionRequiredItem {
```
gcomte, line 280: It's a long name. But hard to shorten, I agree. What do you think about `RequiredAction`?

src/lib.rs
```diff
@@ -1420,6 +1439,35 @@ impl LightningNode {
...
+    /// Returns a list of actionable items. They can be:
+    /// * Uncompleted offers (either available for collection or failed).
+    /// * Unresolved failed swaps.
+    /// * Available funds resulting from channel closes.
+    pub fn list_action_required_items(&self) -> Result<Vec<ActionRequiredItem>> {
+        let uncompleted_offers = self.query_uncompleted_offers()?;
+
+        let failed_swaps = self.get_unresolved_failed_swaps()?;
+
+        let available_channel_close_funds = self.get_node_info()?.onchain_balance;
```
gcomte, line 1453: I'd call that on-chain balance rather than channel close funds, because that's what you are retrieving here. May prevent confusion in the future.

examples/node/cli.rs
danielgranhao: Yes, I would say that how we display things in the CLI is very specific to the CLI itself.

src/lib.rs
danielgranhao: I also thought about that option. It can be misunderstood as an object that represents an action itself instead of something that can be acted upon. We can try to find a better name, but between those two, I prefer my proposal. WDYT?

src/lib.rs
danielgranhao: At least currently, the only way for funds to get to that `onchain_balance` is through channel closes. I proposed this name because we can have other types of onchain funds (i.e., funds from expired swaps), which wouldn't be included here.

examples/node/cli.rs
danielgranhao: Fixed anyway :) [1d9d904](https://github.com/getlipa/lipa-lightning-lib/pull/917/commits/1d9d9045e7db641cd7c1e0c091e582f7394d7136)

src/lib.rs
danielgranhao: I've renamed the variant to be what you propose @andrei-21. [811e841](https://github.com/getlipa/lipa-lightning-lib/pull/917/commits/811e84145e305f38d6f7a02e7e23e7efc534530c)

PR #934, opened 2024-02-15
Release mock artifacts

gcomte:
Whenever there is a new version tag like `v0.38.0-beta`, additional to the regular artifacts, with this PR also a mock artifact is built and deployed to [lipa-lightning-lib-android](https://github.com/getlipa/lipa-lightning-lib-android) resp. [lipa-lightning-lib-swift](https://github.com/getlipa/lipa-lightning-lib-swift) under a new tag, which would in the given case be `v0.38.0-beta-mock`.

### DIFF
--- a/.github/workflows/deploy-release.yml
+++ b/.github/workflows/deploy-release.yml
@@ -211,6 +211,7 @@ jobs:
           git add Package.swift
           git add Sources/LipaLightningLib/LipaLightningLib.swift
           git commit -m "This commit was created automatically by the lipa bot"
+          git pull --rebase
           git tag -a ${{ env.RELEASE_VERSION }}${{ matrix.mock-deps && '-mock' || ''}} HEAD -m "This tag was created automatically by the lipa bot."
           git push
           git push --tag
@@ -314,6 +315,7 @@ jobs:
           git add LipaLightningLib/src/main/java/com/getlipa/lipalightninglib/lipalightninglib.kt
           git add jitpack.yml
           git commit -m "This commit was created automatically by the lipa bot"
+          git pull --rebase
           git tag -a ${{ env.RELEASE_VERSION }}${{ matrix.mock-deps && '-mock' || ''}} HEAD -m "This tag was created automatically by the lipa bot."
           git push
           git push --tag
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-release.yml
```diff
@@ -202,13 +206,15 @@ jobs:
...
           shasum -a 256 lipalightninglibFFI.xcframework.zip | sed 's/ .*//' > checksum
-          sed -i "s/to_replace_release_version/${{ env.RELEASE_VERSION }}/g" Package.swift
+          sed -i "s/to_replace_release_version/${{ env.RELEASE_VERSION }}${{ matrix.mock-deps && '-mock' || '' }}/g" Package.swift
           sed -i "s/to_replace_zip_checksum/$(cat checksum)/g" Package.swift
           
           git add Package.swift
           git add Sources/LipaLightningLib/LipaLightningLib.swift
           git commit -m "This commit was created automatically by the lipa bot"
-          git tag -a ${{ env.RELEASE_VERSION }} HEAD -m "This tag was created automatically by the lipa bot."
+          git restore .
```
danielgranhao, line 215: Why is this restore needed?

danielgranhao (commented):
The `Trigger JitPack build` step should also be updated to use the correct release name.

.github/workflows/deploy-release.yml
gcomte: When we copy around files to create `lipalightninglibFFI.xcframework.zip` before, that creates unstaged changes. With unstaged changes, we cannot `git pull --rebase`. So `git restore .` gets rid of all unstaged changes.

The other question though, is whether `git pull --rebase` is even still needed after I made the to deployments (regular and mock) [run sequentially](https://github.com/getlipa/lipa-lightning-lib/pull/934/commits/645bf9e24aa9018d35cf55fff110d226d472e050). Probably not 🤷
But also it doesn't hurt anyone.

gcomte:
> The `Trigger JitPack build` step should also be updated to use the correct release name.

So [this](7f67c4b3e00e80354eb7eb1f4b01dae0c4e0e7c6) basically, right?

danielgranhao (approved):

PR #963, opened 2024-03-05
Expose synced event

danielgranhao:


tests/print_events_handler/mod.rs
```diff
@@ -22,4 +22,6 @@ impl EventsCallback for PrintEventsHandler {
     fn breez_health_status_changed_to(&self, status: BreezHealthCheckStatus) {
         println!("The Breez SDK health status changed to {status:?}");
     }
+
+    fn synced(&self) {}
```
gcomte, line 26: Why don't you want to print that event?

gcomte (approved):

tests/print_events_handler/mod.rs
danielgranhao: I thought it could result in annoying spam in the CLI. Do you think we should print it?

gcomte: Yes, as discussed over the call.

gcomte (approved):

gcomte (approved):

PR #988, opened 2024-03-20
 Refactor activities interface

andrei-21:


src/lipalightninglib.udl
```diff
@@ -42,7 +42,10 @@ interface LightningNode {
     ListActivitiesResponse get_latest_activities(u32 number_of_activities);
 
     [Throws=LnError]
-    Payment get_payment(string hash);
+    IncomingPaymentInfo get_incoming_payment(string hash);
+
+    [Throws=LnError]
+    OutgoingPaymentInfo get_outgoing_payment(string hash);
```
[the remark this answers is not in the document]
danielgranhao: Also, it would make sense to return entire activities so that in case this is used to fetch e.g. a swap or reverse swap, the swap/reverse-swap specific info is returned as well

src/lib.rs
```diff
@@ -1051,170 +1084,76 @@ impl LightningNode {
...
-        let invoice = parse_invoice(&payment_details.bolt11).map_to_permanent_failure(format!(
-            "Invalid invoice provided by the Breez SDK: {}",
-            payment_details.bolt11
-        ))?;
-        let invoice_details = InvoiceDetails::from_ln_invoice(invoice.clone(), &None);
-
+        let payment_details = match breez_payment.details {
+            PaymentDetails::Ln { ref data } => data,
+            PaymentDetails::ClosedChannel { .. } => {
+                invalid_input!("PaymentInfo cannot be created from channel close")
```
danielgranhao, line 1092: permanent failure from user's PoV

src/lib.rs
```diff
@@ -1517,7 +1445,7 @@ impl LightningNode {
             limit: Some(5),
             offset: None,
         };
-        let latest_payments = self
+        let latest_activity = self
```
danielgranhao, line 1448:
```suggestion
        let latest_activities = self
```

examples/node/cli.rs
```diff
@@ -1123,59 +1124,91 @@ fn list_activities(node: &LightningNode, words: &mut dyn Iterator<Item = &str>)
...
-        "      LSP fees:         {:?}",
-        payment.lsp_fees.map(|f| amount_to_string(&f))
+        "      LSP fees:         {}",
+        amount_to_string(&payment.lsp_fees),
     );
-    println!("      Hash:             {}", payment.hash);
-    println!("      Preimage:         {:?}", payment.preimage);
-    println!("      Description:      {}", payment.description);
+    //    println!("      Offer:            {}", offer_to_string(payment.offer));
+    //    println!("      Swap:             {:?}", payment.swap);
```
danielgranhao, line 1195: Why commented?

danielgranhao (approved):

PR #1019, opened 2024-04-12
Harmonize mock payment creation

gcomte:
Mostly a refactoring, added some dummy data in a couple of places

### DIFF
--- a/mock/breez-sdk/src/lib.rs
+++ b/mock/breez-sdk/src/lib.rs
@@ -165,36 +165,22 @@ impl BreezServices {
                     *LN_BALANCE_MSAT.lock().unwrap() -= amount_msat;
                 }
 
-                let payment = Payment {
-                    id: "".to_string(),
+                let payment = create_payment(MockPayment {
                     payment_type: PaymentType::Sent,
-                    payment_time: Utc::now().timestamp(),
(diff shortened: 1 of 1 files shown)


mock/breez-sdk/src/lib.rs
```diff
@@ -1146,6 +1044,58 @@ fn generate_2_hashes_raw() -> (sha256::Hash, sha256::Hash) {
...
+fn create_payment(p: MockPayment) -> Payment {
+    let now = Utc::now().timestamp();
+
+    Payment {
+        id: now.to_string(), // Placeholder. ID is probably never used
+        payment_type: p.payment_type,
+        payment_time: now,
+        amount_msat: p.amount_msat,
+        fee_msat: p.fee_msat,
+        status: PaymentStatus::Complete,
```
danielgranhao, line 1073: Do we never need to create pending payments?

gcomte: Yeah maybe we do. Might be a future requirement or even bugfix.

However, for this refactoring I just checked what values are being used and the current code always uses `PaymentStatus::Complete` everyhwere, so that's why I went with that.

danielgranhao (approved):

PR #1024, opened 2024-04-16
Prevent double sync in Mock

gcomte:
DRAFT: I first need to finish the git rebase for this to make sense


gcomte:
The sync is needed because the other one happens too late.

PR #1028, opened 2024-04-16
Prevent DeadLock

gcomte:
Otherwise the lock is being kept for the entire `push` operation within which also `get_exchange_rate` tries to acquire it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -973,12 +973,14 @@ impl LightningNode {
                 "Failed to get in-progress swap",
             )?
         {
+            let created_at = unix_timestamp_to_system_time(in_progress_swap.created_at as u64)
+                .with_timezone(self.user_preferences.lock_unwrap().clone().timezone_config);
+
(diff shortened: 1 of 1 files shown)


danielgranhao (approved):

src/lib.rs
```diff
@@ -973,12 +973,15 @@ impl LightningNode {
                 "Failed to get in-progress swap",
             )?
         {
+            let created_at = unix_timestamp_to_system_time(in_progress_swap.created_at as u64)
+                .with_timezone(self.user_preferences.lock_unwrap().clone().timezone_config);
+
+            // Convert UTC time to a specific time zone
```
danielgranhao, line 979:
```suggestion
            // Convert UTC time to a specific time zone
            let created_at = unix_timestamp_to_system_time(in_progress_swap.created_at as u64)
                .with_timezone(self.user_preferences.lock_unwrap().clone().timezone_config);
```

PR #1038, opened 2024-04-22
Expose LNURL-pay comment

andrei-21:


examples/node/cli.rs
```diff
@@ -1213,6 +1213,10 @@ fn print_outgoing_payment(payment: OutgoingPaymentInfo) -> Result<()> {
         amount_to_string(&payment.network_fees)
     );
     println!("      Recipient:        {:?}", payment.recipient);
+    println!(
+        "   Comment sent:        {:?}",
```
danielgranhao, line 1217:
```suggestion
        "      Comment sent:   {:?}",
```

src/lipalightninglib.udl
```diff
@@ -362,6 +362,7 @@ dictionary OutgoingPaymentInfo {
     PaymentInfo payment_info;
     Amount network_fees;
     Recipient recipient;
+    string? comment_for_recipient;
```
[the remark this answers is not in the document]
danielgranhao: I think it's okay. An alternative I see is to use another enum instead of an optional value, but I'm not sure it's worth it in this case.

src/payment.rs
```diff
@@ -184,6 +184,8 @@ pub struct OutgoingPaymentInfo {
     pub network_fees: Amount,
     /// Information about a payment's recipient.
     pub recipient: Recipient,
+    /// Comment sent to the recipient.
```
danielgranhao, line 187:
```suggestion
    /// Comment sent to the recipient. Only set for LNURL payments where a comment has been sent.
```

danielgranhao (approved):

PR #1053, opened 2024-05-06
Remove `RecommendedAction`

danielgranhao:


src/notification_handling.rs
```diff
@@ -154,36 +149,36 @@ fn handle_payment_received_notification(
...
 
     // Wait for payment to be received
     if let Some(details) = wait_for_payment_with_timeout(event_receiver, &payment_hash)? {
-        return Ok(RecommendedAction::ShowNotification {
-            notification: Notification::Bolt11PaymentReceived {
-                amount_sat: details.payment.map(|p| p.amount_msat).unwrap_or(0) / 1000, // payment will only be None for corrupted GL payments. This is unlikely, so giving an optional amount seems overkill.
-                payment_hash,
-            },
+        return Ok(Notification::Bolt11PaymentReceived {
+            amount_sat: details.payment.map(|p| p.amount_msat).unwrap_or(0) / 1000, // payment will only be None for corrupted GL payments. This is unlikely, so giving an optional amount seems overkill.
```
gcomte, line 164: Yes not an `Option`, but why returning `Ok()` with an `amount_sat` of `0` rather than just returning an error?

What is a _corrupted GL_ payment even? Does it mean the payment did actually go through, but GL cannot provide data for it? Then too I'm not sure whether it is better to display a wrong payment or just no payment at all.

src/notification_handling.rs
```diff
@@ -216,25 +211,25 @@ fn handle_address_txs_confirmed_notification(
...
 
     // Wait for payment to arrive
     if let Some(details) = wait_for_payment_with_timeout(event_receiver, &payment_hash)? {
-        return Ok(RecommendedAction::ShowNotification {
-            notification: Notification::OnchainPaymentSwappedIn {
-                amount_sat: details.payment.map(|p| p.amount_msat).unwrap_or(0) / 1000, // payment will only be None for corrupted GL payments. This is unlikely, so giving an optional amount seems overkill.
-                payment_hash,
-            },
+        return Ok(Notification::OnchainPaymentSwappedIn {
+            amount_sat: details.payment.map(|p| p.amount_msat).unwrap_or(0) / 1000, // payment will only be None for corrupted GL payments. This is unlikely, so giving an optional amount seems overkill.
```
gcomte, line 223: Same here

src/lipalightninglib.udl
```diff
@@ -770,4 +764,5 @@ interface NotificationHandlingError {
 enum NotificationHandlingErrorCode {
     "NodeUnavailable",
     "InProgressSwapNotFound",
+    "ExpectedPaymentNotReceived",
```
gcomte, line 767: When looking at the implementation this error code makes a lot of sense, but from the PoV of the library consumer, I think this error code could be very confusing.

Wouldn't something like this be more telling? --> `CorruptPaymentNotification`

Also for the `InProgressSwapNotFound` --> `CorruptInProgressSwapNotification`

WDYT?

src/errors.rs
```diff
@@ -304,6 +304,8 @@ pub enum NotificationHandlingErrorCode {
     NodeUnavailable,
     /// The specificed in-progress swap couldn't be found.
     InProgressSwapNotFound,
+    /// A payment was expected, but it was not received.
```
gcomte, line 307: Is this comment necessary (for the docs)?
It doesn't add any information that cannot already be derived from the enum variants name.

On the other hand, you could actually provide some clarification here imo:
_Inconsistent payment data: A payment notification has been received, but the corresponding payment could not be retrieved from the node._

(Same is true for the other variants)

src/notification_handling.rs
danielgranhao: > What is a corrupted GL payment even? Does it mean the payment did actually go through, but GL cannot provide data for it?

Yes, the payment goes through.

It's likely better to return information communicating that the expected payment was received, even if the returned amount is incorrect. If we want to capture the possibility of these corrupted payments correctly, an optional value would be best, but I don't think we should add that complexity to the interface to accommodate such an unlikely event.

src/lipalightninglib.udl
danielgranhao: I don't think the notifications were necessarily corrupt. `ExpectedPaymentNotReceived ` will likely be caused by an issue with the node or LSP that prevented the payment from being received. `InProgressSwapNotFound` may be caused by another client running concurrently and processing the swap first. 

I think it's better to keep error names descriptive of what occurs, but I may be missing the library consumer perspective. 

In the end, given how tightly coupled we are with the library consumer, either approach is fine. Let me know if you strongly prefer other names.

gcomte (approved):

src/errors.rs
danielgranhao: Docs improved in [6569e5d](https://github.com/getlipa/lipa-lightning-lib/pull/1053/commits/6569e5df976161a85fd6b62e277bbc67201b5815)

PR #1054, opened 2024-05-06
Document network usage in methods

danielgranhao:


gcomte (approved):
Didn't verify the actual network usage of each function, but I love the idea of documenting it.

Would be nice to have automated regression testing that verify the continuous accuracy of that documentation.

danielgranhao:
> Didn't verify the actual network usage of each function, but I love the idea of documenting it.
> 
> Would be nice to have automated regression testing that verify the continuous accuracy of that documentation.

I'm quite confident I got it right, but given the sheer amount of methods, maybe I got something wrong. Anyway, we can fix it later if/when we detect it. The consequences of an error here are small.

Interesting idea. Do you have a suggestion for how that automated verification could be implemented?

PR #1055, opened 2024-05-06
Temporarilly disable cargo audit

danielgranhao:


gcomte:
Is there anything we did to promote the fixing of our affected dependencies?

danielgranhao:
> Is there anything we did to promote the fixing of our affected dependencies?

The fix will likely come without us asking for it. Dependabot will inform us about updates. We will also not forget about this skipped job because every PR shows that it was skipped. If we don't see this fixed for a while, we can create an issue in Breez SDK.

gcomte (approved):

PR #1067, opened 2024-05-14
Implement mocking of channels

danielgranhao:


mock/breez-sdk/src/lib.rs
```diff
@@ -808,8 +859,8 @@ impl BreezServices {
         let now = Utc::now().timestamp();
 
         let amount_msat = req.amount_sat * 1_000;
-        let fee_msat = *LN_BALANCE_MSAT.lock().unwrap() - amount_msat;
-        *LN_BALANCE_MSAT.lock().unwrap() -= amount_msat + fee_msat;
+
+        send_payment_mock_channels(amount_msat + 1234);
```
gcomte, line 863: The fee `1234 msat` is being used in 3 places, maybe let's make it a constant for easier understanding

mock/breez-sdk/src/lib.rs
```diff
@@ -1041,7 +1096,6 @@ impl BreezServices {
                 reverse_swap_info: None,
             });
 
-            *LN_BALANCE_MSAT.lock().unwrap() -= amount_msat;
```
gcomte, line 1044: Yeah okay I see, it's not strictly necessary ... But why did you remove this? I thought it would keep the wallet coherent

gcomte (commented):
nice!

mock/breez-sdk/src/lib.rs
danielgranhao: Do you mean the total balance variable? We now use channels as state. Having the variable introduced information duplication and forces us always to update both parts in all the places

mock/breez-sdk/src/lib.rs
danielgranhao: I replaced it with randomly generated values instead :)

mock/breez-sdk/src/lib.rs
danielgranhao: We clarified in a side-channel that @gcomte was referring to the mismatch between the balance and payments that is created by my changes to `simulate_payments()`. We nonetheless agreed the mismatch is not important so no changes are needed.

gcomte (approved):

PR #1071, opened 2024-05-15
Update Breez SDK version in Breez SDK mock

gcomte:


danielgranhao (approved):
Thanks, forgot to do this

PR #1082, opened 2024-05-20
Extract Pocket client to its own crate

danielgranhao:


danielgranhao:
> The topup int test isn't successful anymore. I reran it but it still fails.

@gcomte the test is failing in all CI runs, not just from this PR. Seems to be an issue with the service. I've reported it already.

gcomte (approved):

PR #1166, opened 2024-08-14
Use accept terms and conditions v2

danielgranhao:


gcomte (approved):
Don't forget to update the Cargo.toml reference after releasing a new wild version of course

PR #1190, opened 2024-09-17
Add E2E tests to periodically test GL/Breez/Backend-integration

gcomte:


tests/sat_flows.rs
```diff
@@ -0,0 +1,308 @@
...
+        if sender_sent_payment && receiver_received_payment {
+            return Ok(());
+        }
+
+        std::thread::sleep(std::time::Duration::from_secs(1));
+    }
+}
+
+fn get_payment_amount() -> PaymentAmount {
+    let exact = 1 + rand::random::<u64>() % MAX_PAYMENT_AMOUNT_SATS;
```
danielgranhao, line 300: I am unsure if there is a reason to test with different amounts, but if there is, we shouldn't use random ones (we add variability between tests). One option would be to always run 3 payments (e.g. 50, 500, 5000 sats). But always doing 5000 seems enough.

tests/sat_flows.rs
```diff
@@ -0,0 +1,308 @@
...
+        node,
+        received_payments,
+        sent_payments,
+    }
+}
+
+fn wait_for_payment(
+    payment_hash: &str,
+    sender: &Arc<Mutex<Vec<String>>>,
+    receiver: &Arc<Mutex<Vec<String>>>,
```
danielgranhao, line 265: You could use channels here instead, but this works (just not as pretty 😃 )

.github/workflows/e2e_tests.yml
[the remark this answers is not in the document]
gcomte: It's dirt cheap

.cargo/config.toml.breez.sample
```diff
@@ -30,5 +30,8 @@ BREEZ_SDK_API_KEY = ""
 BREEZ_SDK_PARTNER_CERTIFICATE = ""
 BREEZ_SDK_PARTNER_KEY = ""
 
-BREEZ_SDK_MNEMONIC_ALICE = ""
-BREEZ_SDK_MNEMONIC_BOB = ""
+BREEZ_SDK_MNEMONIC_EMPTY = ""
+BREEZ_SDK_MNEMONIC_FUNDED_SENDER = ""
+BREEZ_SDK_MNEMONIC_FUNDED_RECEIVER = ""
+TESTING_RECEIVER_LN_ADDRESS = ""
+TESTING_RECEIVER_PHONE_NUMBER = ""
```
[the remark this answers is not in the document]
gcomte: No, they are not being used. Thanks for pointing out. [Fix](https://github.com/getlipa/lipa-lightning-lib/pull/1190/commits/3dac2f2ad4ac34ee25577a861d18eaa74d45c681)

tests/sat_flows.rs
gcomte: [I agree](https://github.com/getlipa/lipa-lightning-lib/pull/1190/commits/d487ae5a052872d163b67105e882d5d29ff020ac), just not sure about the sats amount. I set it to 300 SAT to save some fees. What should we base our `golden number` on?

danielgranhao: For our main purpose with this, the amount shouldn't matter much. 300 is fine by me :)

danielgranhao (approved):

tests/sat_flows.rs
gcomte: Thanks for the hint: https://github.com/getlipa/lipa-lightning-lib/pull/1190/commits/c04a8dc51f4522b2cad0391bd211db43d13cfe82

danielgranhao (approved):

PR #1228, opened 2024-10-21
Print env var key when value is not present

gcomte:
Int tests seem to fail because of a missing environment variable, but it is unclear which one is missing.


tests/setup/mod.rs
```diff
@@ -93,7 +93,7 @@ pub fn start_specific_node(
     }
 
     let mnemonic_key = format!("BREEZ_SDK_MNEMONIC_{node_name}");
-    let mnemonic = std::env::var(mnemonic_key).unwrap();
+    let mnemonic = std::env::var(mnemonic_key).expect("Failed to load env var {mnemnoic_key}");
```
danielgranhao, line 96: You should use `format!` otherwise you'll just print `{mnemonic_key}`. I only noticed because you have a typo on the var name (`mnemoic_key`) but this was still built fine 😆

danielgranhao (changes requested):

tests/setup/mod.rs
gcomte: lol, true

danielgranhao (approved):

PR #1273, opened 2024-11-20
Use longer lived lsp fee offers for redeeming

gcomte:
An LSP may offer multiple fee rates, tied to different expiration dates.

This PR exposes this election offer-election in some places and uses opinionated values in others.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -638,13 +638,8 @@ impl LightningNode {
     ///
     /// Requires network: **yes**
     #[deprecated = "lightning().calculate_lsp_fee_for_amount() should be used instead"]
-    pub fn calculate_lsp_fee(
-        &self,
-        amount_sat: u64,
-        expiry: Option<u32>,
-    ) -> Result<CalculateLspFeeResponse> {
-        self.lightning
-            .calculate_lsp_fee_for_amount(amount_sat, expiry)
+    pub fn calculate_lsp_fee(&self, amount_sat: u64) -> Result<CalculateLspFeeResponse> {
+        self.lightning.calculate_lsp_fee_for_amount(amount_sat)
     }
 
     /// Get the current limits for the amount that can be transferred in a single payment.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -625,13 +625,9 @@ impl LightningNode {
 
     /// Calculate the actual LSP fee for the given amount of an incoming payment.
     /// If the already existing inbound capacity is enough, no new channel is required.
-    /// The LSP may offer multiple fee rates, tied to different expiration dates.
-    /// Increased expiry dates mean higher fee rates.
-    /// This method returns the best offer within the given expiry.
     ///
     /// Parameters:
     /// * `amount_sat` - amount in sats to compute LSP fee for
-    /// * `expiry` - expiry time in seconds
     ///
     /// For the returned fees to be guaranteed to be accurate, the returned `lsp_fee_params` must be
     /// provided to [`LightningNode::create_invoice`]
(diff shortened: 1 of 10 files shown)


src/lipalightninglib.udl
```diff
@@ -326,6 +326,9 @@ interface Swap {
     [Throws=SwapError]
     SwapAddressInfo create(OpeningFeeParams? lsp_fee_params);
 
+    [Throws=LnError]
+    CalculateLspFeeResponse calculate_swap_lsp_fee_for_amount(u64 amount_sat);
```
danielgranhao, line 330: Given we are within the swap interface, there is no need to have "swap" in the method name :) 

```suggestion
    CalculateLspFeeResponse calculate_lsp_fee_for_amount(u64 amount_sat);
```

src/lib.rs
```diff
@@ -625,9 +625,13 @@ impl LightningNode {
 
     /// Calculate the actual LSP fee for the given amount of an incoming payment.
     /// If the already existing inbound capacity is enough, no new channel is required.
+    /// The LSP may offer multiple fee rates, tied to different expiration dates.
+    /// Increased expiry dates mean higher fee rates.
+    /// This method returns the best offer within the given expiry.
     ///
     /// Parameters:
     /// * `amount_sat` - amount in sats to compute LSP fee for
+    /// * `expiry` - expiry time in seconds
```
danielgranhao, line 634: The docs here are not up-to-date. We probably don't want any changes

src/onchain/channel_closes.rs
```diff
@@ -239,7 +239,7 @@ impl ChannelClose {
 
         let lsp_fees = self
             .support
-            .calculate_lsp_fee_for_amount(send_amount_sats)
+            .calculate_lsp_fee_for_amount(send_amount_sats, None)
```
danielgranhao, line 242: This logic needs to be fixed anyway (we should use lsp fee params provided as input -  let's fix in another PR), but I would use here the new swap version of the method.

src/onchain/swap.rs
```diff
@@ -232,7 +234,7 @@ impl Swap {
 
         let lsp_fees = self
             .support
-            .calculate_lsp_fee_for_amount(send_amount_sats)?
+            .calculate_lsp_fee_for_amount(send_amount_sats, None)?
```
danielgranhao, line 237: This logic needs to be fixed anyway (we should use lsp fee params provided as input -  let's fix in another PR), but I would use here the new swap version of the method.

src/onchain/swap.rs
```diff
@@ -282,6 +284,27 @@ impl Swap {
...
+    ) -> Result<CalculateLspFeeResponse> {
+        calculate_swap_lsp_fee_for_amount(&self.support, amount_sat)
+    }
+}
+
+pub fn calculate_swap_lsp_fee_for_amount(
+    support: &Support,
+    amount_sat: u64,
+) -> Result<CalculateLspFeeResponse> {
+    support.calculate_lsp_fee_for_amount(amount_sat, Some(TWO_WEEKS))
```
danielgranhao, line 307: Do we need this function? It seems to me that everywhere it is needed we have an instance of `Swap`, so we can simply use the method

src/lib.rs
gcomte: Removed: 2a522c732a12c16c00c216b17e7199e9199a1819

src/onchain/channel_closes.rs
gcomte: I'm gonna create a second PR :+1:

gcomte: Temporary fix: 9854933355983b8f7fc790547759b225ce1dc4eb

src/onchain/swap.rs
gcomte: Not really. But I introduced `Swap` where it was missing: 5e420e731955c7579898a75f26761c19658991a0

src/lipalightninglib.udl
gcomte: d7e25c0e7143bb8e9864d23e47f4ea1804d1bd46

PR #1283, opened 2024-11-28
Implement configurable external recovery

danielgranhao:


src/lib.rs
```diff
@@ -5,6 +5,7 @@
 //! Most functionality can be accessed by creating an instance of [`LightningNode`] and using its methods.
 
 #![allow(clippy::let_unit_value)]
+#![allow(clippy::empty_line_after_doc_comments)]
```
gcomte, line 8: Either we do this, or the [already merged change](https://github.com/getlipa/lipa-lightning-lib/pull/1284), but better not both

src/backup.rs
```diff
@@ -46,15 +46,23 @@ impl BackupManager {
             .map_to_permanent_failure("Failed to perform backup of local db")
     }
 
-    pub async fn recover(&self) -> Result<()> {
+    pub async fn recover(&self, external_recovery_is_allowed: bool) -> Result<()> {
```
gcomte, line 49: Why not just `allow_external_recovery`?

src/lib.rs
danielgranhao: Let's use my solution. Changing the CI toolchain version does not fix the issue on local environments. 

I've added a comment explaining why this is done.

src/backup.rs
danielgranhao: Applied in [69bd723](https://github.com/getlipa/lipa-lightning-lib/pull/1283/commits/69bd72318f569863776ccb56791c5a9c009528aa)

gcomte (approved):

PR #1310, opened 2024-12-18
Control on-chain balance on `mimic.activities`

andrei-21:
Such that we can test small amounts which are impossible to sweep


gcomte (approved):

gcomte:
ACK but needs to be well explained in the docs, as usually we don't use the amount for any command

PR #1343, opened 2025-01-21
Update deprecated set state commands for GH actions

gcomte:
See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


gcomte:
After merging a test release should be crafted.

dleutenegger (approved):

